PR #9, opened 2024-05-28
Dynamically generate the year variable

sebbASF:
This will help maintain compatibility with existing Pelican processing


gstein:
Thanks!

PR #22, opened 2024-06-01
Use the plugins that are already available to the action

sebbASF:
When an action is used, its code is loaded onto the host, and is available for use by the action.
This means that an action can directly use scripts and data files, for example the plugins.
There is no need to fetch the files.

Instead of using fetch_plugins.py to overwrite the site plugins directory, the new plugin_paths.py script is used to generate a suitable override value for PLUGIN_PATHS which is passed as a CLI argument to Pelican.

### DIFF
--- a/pelican/action.yml
+++ b/pelican/action.yml
@@ -58,7 +58,7 @@ runs:
           else
             exec >/dev/null
           fi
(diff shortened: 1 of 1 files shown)


gstein:
Very cool! Dramatic change/simplify

dfoulks1 (commented):
lgtm

PR #32, opened 2024-06-05
Add stash action

assignUser:
The stash action is an alternative to actions/cache that doesn't suffer it's issues for use of storing build cache artifacts like ccache_dir etc.. Details in the readme.

This action was previously developed in https://github.com/assignUser/stash under ASL 2.0 and I am the sole contributor so there should be no  copyright or license issues moving this here.  

Once this PR is merged I will make the previous defunct and print out a message about this repo and then archive it.


assignUser:
I was waiting for this repo to exists so we can use the action in apache/arrow* but it's currently in heavy use in [Velox](https://github.com/facebookincubator/velox/actions/runs/9388518675/job/25853929452#step:5:1) where it's used to store ccache caches. Previously we used actions/cache in Velox which was not very effective as the 10GB/repo limit for caches meant that cache evictions and there for misses where plenty.

assignUser:
Thanks for the review.
> I suggest waiting for their go-ahead before spending time updating the PR to address these comments

We talked about this at the round table today :)

assignUser:
Sorry for my lack of response, was a bit busy but thank you for your continued review, I will get to it soon.

> How does this action differ from actions/upload-artifact and actions/download-artifact?

The main difference is that this action emulates the scoping rules of `actions/cache` and prevents untrusted user caches (e.g. from fork PR CI runs) from being used on the default branch (where actions have elevated permissions by default) or  other repo or PR branches. This is a security feature to prevent cache poisoning attacks. The reason I use python instead of bash is to prevent code injection through `${{ }}` in envvars/bash directly that would potentially allow circumventing these scoping rules.

assignUser:
Wow, it took me way to long to get back to this, sorry.

PR #51, opened 2024-06-11
Use artifacts to cache the build

sebbASF:
The PR looks for a GFM build artifact in the repo housing the action.
If one is found, it is downloaded and the build is skipped.

It can take 15+ seconds to build GFM, which is large proportion of the run-time, so it seems worth doing.

[A previous version of this PR also checked the calling repo for the artifact, and would create an artifact if none was found there either. However this would potentially create artifacts in lots of repos, as well as being an unexpected side effect of the build action]


gstein:
Should we create a distinct action for this? eg. infrastructure-actions/build-gfm ? ... It seems there is a fair bit of logic around this, which could be pulled out into a separate action for clarity of maintenance. ... I don't have a particular opinion either way. The current action.yml isn't so large as to be confusing.

All that said, I do think that caching the artifact is the right approach. No need to keep rebuilding it for $N sites and runs.

gstein:
Oh. I just saw your reference to #56 ... but my query is: why duplicate build logic between that and this one? And #56 has a very different looking code for a build (it has a lot of cron/expiration checks) ... could we roll up *all* logic around building/stashing the artifact into a new Action. ?

gstein:
build-cmark.sh is fair game, as it is only used within this repository/action. Use your best judgement. Thanks!

gstein:
Offhand, let the Action define the LIBCMARKDIR, and hand that to build-cmark.sh. Is that your thinking?

gstein:
At a minimum, build-cmark.sh could simply `mv cmark.so $LIBDIR_GIVEN_TO_US`

pelican/action.yml
```diff
@@ -51,34 +51,55 @@ runs:
...
+        # Is there a saved artifact for the GFM build?
+        echo "Check for GFM build artifact in action repo: $GITHUB_ACTION_REPO"
+        gh run download --dir ${LIBCMARKDIR} --name ${GFM_ARTIFACT_KEY} --repo $GITHUB_ACTION_REPO || true
+        if [[ -f $LIBCMARKDIR/libcmark-gfm.so ]]
+        then
+          echo "Downloaded to ${LIBCMARKDIR} from $GITHUB_ACTION_REPO, nothing more to do!"
+          exit 0 # nothing more to do in this step
+        fi
+
+        # GFM binary not found, need to build it
```
assignUser, line 91: In contrast to caching compile caches (see #32 ) this is actually a good/intended use case for https://github.com/actions/ cache because you have the fixed key of the gfm version and don't need any incremental updates.

The cache action also handles scoping of the cache so that a run of the action on the default branch won't pull in an artifact from a (user supplied, thus untrusted) fork PR CI run. You can then simply do this in an earlier step:

```
- name: Cache GFM
  id: gfm-cache
  uses: actions/cache@v4
  with:
    path: ${{ env.LIBCMARKDIR }}
    key: gfm-lib-${{ env.GFM_VERSION }}
```
and use the output `steps.gfm-cache.outputs.cache-hit` in an `${{ }}` expression to check if the lib was restored, if it was not the action will automatically cache `path:` at the end of the action.

assignUser: (some people recommend pinning even `actions/*` actions to specific shas in jobs with elevated permissions but that is currently not required by the gha policy)

assignUser: Sorry for the notification spam. Is pelican not compatible with `libcmark-gfm 0.29.0` which is available via `apt` in both 20.04 and 22.04?

assignUser:
>Also, changing from BuildBot to GH CI is a big change, and the fewer other changes that are made, the easier it is to debug problems. Updates to GFM and Pelican versions can be made later.

:+1:  makes sense, was just wondering :)

> whereas the cached artifacts are always available to new branches and new repos.

True but they also add a considerable amount of code/process to maintain and with the build only taking 15 seconds I'd lean towards 'no code is the best code' ^^ If the build was more substantial, this would be a great setup, but for 15 seconds it seems a bit overkill. An alternative would be to use a docker image in the step to run pelican, or even a complete docker action but I don't really have any experience with Pelican so :shrug: ghcr.io caching + a minimized image makes that pretty fast too.

> not sure if it helps with the first build when a new branch is cloned

If there is a cache on the base branch then yes.

assignUser:
> I tried using a Docker image and that was slower.

To clarify before I go down a rabbit hole to test this: you used the docker file and build the image in the workflow or  pre-built a docker image and hosted it on ghcr.io and used that in the workflow?

potiuk:

PR #52, opened 2024-06-12
Allow default TOC container format to be overridden

sebbASF:
Allow the TOC container format to be overridden so toc.py can generate the same output as toc2.py.

If implemented, toc2.py can be dropped, provided that any pelicanconf.py settings files are updated accordingly.


dfoulks1:
LGTM, but someone with more familiarity with TOC should also take a look

dfoulks1:
I'm going to merge this, if there's an issue we can always roll back.

PR #53, opened 2024-06-14
Extract artifact key name

sebbASF:


pelican/action.yml
```diff
@@ -59,13 +59,15 @@ runs:
...
         GH_TOKEN: ${{ github.token }} # needed by gh
       run: |
+        export GFM_ARTIFACT_KEY=cmark-gfm-${GFM_VERSION}
         # Needs to be available to a subsequent step
         echo  "GFM_VERSION=${GFM_VERSION}" >>$GITHUB_ENV
+        echo  "GFM_ARTIFACT_KEY=${GFM_ARTIFACT_KEY}" >>$GITHUB_ENV
         if [[ -z $LIBCMARKDIR ]] # don't overwrite LIBCMARKDIR if it exists
         then
           # set up the GFM environment
           export LIBCMARKDIR=${WORKDIR}/cmark-gfm-${GFM_VERSION}/lib
```
gstein, line 69: note that GFM_ARTIFACT_KEY could be used here.

but that also shows it isn't just an artifact key. It is the name of the library, at its heart. (which also happens to be used as a key, too) ... maybe GFM_LIBNAME ?

gstein: 👍

PR #56, opened 2024-06-15
Workflow to create GFM artifact(s)

sebbASF:
This workflow runs once a week to create the GFM artifact if necessary.
Most weeks it should do nothing, as there should already be an artifact present.
The gap between runs could be increased, provided that the expiry date checking was also updated.


.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,64 @@
...
+
+        # Check if artifact is present (list all)
+        curl -sS https://api.github.com/repos/sebbASF/infra-actions/actions/artifacts?name=${GFM_ARTIFACT_KEY} >/tmp/artifact.json
+        # when does last one expire?
+        jq </tmp/artifact.json '[.artifacts[]|select(.expired==false)|.expires_at]|max' >/tmp/max.txt
+        # is that more than 10 days away? (86400*10 seconds)
+        OUT=$(jq </tmp/max.txt 'select(.>(now+864000|strftime("%FT%TZ")))')
+        if [[ -n $OUT ]]
+        then
+          echo "Found a valid artifact for ${{ matrix.gfm_version }} (expires $OUT)"
```
jsoref, line 47: you should never use `${{ ... }}` inside a `run:` block.

use:
```yaml
env:
  gfm_version: ${{ matrix.gfm_version }}
...
run:
          echo "Found a valid artifact for $gfm_version (expires $OUT)"
...
```

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,64 @@
...
+        mkdir -p ${LIBCMARKDIR}
+        bash $GITHUB_WORKSPACE/pelican/build-cmark.sh ${{ matrix.gfm_version }} ${LIBCMARKDIR}
+
+        # Tell the save step what to save
+        echo "created=${LIBCMARKDIR}" | tee -a $GITHUB_OUTPUT
+    - name: Save the GFM build ${{ matrix.gfm_version}}
+      if: ${{ steps.build_gfm.outputs.created }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ env.GFM_ARTIFACT_KEY}}
```
jsoref, line 63:
```suggestion
        name: ${{ env.GFM_ARTIFACT_KEY }}
```

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,64 @@
...
+
+        echo "Could not find a valid artifact for ${{ matrix.gfm_version }}; building another"
+        # build GFM and set up LIBCMARKDIR
+        export LIBCMARKDIR=/tmp/gfm-${{ matrix.gfm_version }}
+        mkdir -p ${LIBCMARKDIR}
+        bash $GITHUB_WORKSPACE/pelican/build-cmark.sh ${{ matrix.gfm_version }} ${LIBCMARKDIR}
+
+        # Tell the save step what to save
+        echo "created=${LIBCMARKDIR}" | tee -a $GITHUB_OUTPUT
+    - name: Save the GFM build ${{ matrix.gfm_version}}
```
jsoref, line 59:
```suggestion
    - name: Save the GFM build ${{ matrix.gfm_version }}
```

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,64 @@
...
+      - 'pelican/build-cmark.sh'
+  workflow_dispatch:
+
+jobs:
+  build-artifact:
+    runs-on: ubuntu-latest
+    strategy:
+      # Allow for multiple versions to be maintained
+      matrix:
+        gfm_version: ['0.28.3.gfm.12']
```
jsoref, line 27:
```suggestion
        gfm_version:
        - '0.28.3.gfm.12'
```

Please use some form of the array notation so that when people add / remove items they aren't blamed for changing everything.

.github/workflows/build-artifacts.yml
jsoref: Depending on what you're doing, it can result in security vulnerabilities or very surprising misbehaviors.

In this case, if your matrix was:
```yaml
matrix:
  gfm_version:
  - 'hello"; touch /tmp/flag; echo "'
```

then you'd be constructing a script of the form:
```sh
          echo "Found a valid artifact for hello"; touch /tmp/flag; echo " (expires $OUT)"
```

which is fairly surprising.
https://github.com/check-spelling-sandbox/sturdy-adventure/actions/runs/13166687061/job/36748420310#step:6:14

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,64 @@
...
+        persist-credentials: false
+    - name: Fetch and build version ${{ matrix.gfm_version }}
+      id: build_gfm
+      run: |
+        # N.B. This must agree with the definition in pelican/action.yml
+        export GFM_ARTIFACT_KEY=gfm-lib-${{ matrix.gfm_version }}
+        echo "GFM_ARTIFACT_KEY=${GFM_ARTIFACT_KEY}" >> $GITHUB_ENV
+
+        # Check if artifact is present (list all)
+        curl -sS https://api.github.com/repos/sebbASF/infra-actions/actions/artifacts?name=${GFM_ARTIFACT_KEY} >/tmp/artifact.json
```
jsoref, line 40: Oh, this probably shouldn't be `sebbASF`

You should use `$GITHUB_REPOSITORY`:
```suggestion
        curl -sS https://api.github.com/repos/$GITHUB_REPOSITORY/actions/artifacts?name=${GFM_ARTIFACT_KEY} >/tmp/artifact.json
```

pelican/action.yml
```diff
@@ -51,34 +51,55 @@ runs:
...
+        then
+          echo "Downloaded to ${LIBCMARKDIR} from $GITHUB_ACTION_REPO, nothing more to do!"
+          exit 0 # nothing more to do in this step
+        fi
+
+        # GFM binary not found, need to build it
         {
+          echo "Creating GFM binary in ${LIBCMARKDIR}"
           # disable stdout unless debug is on
           if [ "${{ inputs.debug }}" == 'true' ]
```
jsoref, line 95: it's nice if you automatically debug on rerun...

```
|| [ "$GITHUB_RUN_ATTEMPT" != 1 ]
```

jsoref: Note that you should also move this to an `env:`, it isn't just `${{ matrix`, it's `${{` _anything_

jsoref: for most of those, github has a $GITHUB_..., so you really should never do it :)... it's a safe enough rule.

potiuk:

PR #63, opened 2024-06-20
INFRA-25890 Pelican Action: Add support for installing requirements.txt

janhoy:
Some web sites have custom Pelican plugins that require python modules. This PR will accept a new input `requirements` where user can specify name of a file to load python requirements from. Default is `requirements.txt`. If the specified file does not exist, we'll log a line and continue the build.

https://issues.apache.org/jira/browse/INFRA-25890


janhoy:
@sebbASF , @dfoulks1 This problem is needed for the Solr project website to re-enable a plugin we have been using. Appreciate a review.

PR #67, opened 2024-06-26
Requirments.txt should be truly optional

sebbASF:
Don't complain if requirements.txt does not exist; it should be optional.


gstein:
Thanks!

PR #76, opened 2024-10-17
Conditional BSP

sebbASF:
Pip will only take note of this if it is supported (props stack overflow)


dave2wave (commented):
Wouldn't a better approach be to switch to using venv like Ubuntu 24.04 prefers?

PR #77, opened 2024-12-03
Dfoulks/asciidoc

dfoulks1:
Requesting to add the AsciiDoc action (builds Asciidoc files and commits to perscribed branch)

## Uses 
* checkout@v3

## Process
* Checks out code
* Installs nodejs + asciidoc packages
* iterates through adoc files in `$srcfiles`
* moves HTML and `$assets` to tempdir.
* checks out new branch
* cleans out cruft
* commits new stuff to `$dest branch`


asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+    - uses: actions/setup-node@v4
+    - name: Install AsciiDoc
+      shell: bash
+      # Install needs to run in separate shell so stdout is restored
+      run: |
+        (
+          test "${{ inputs.debug }}" == 'true' || exec >/dev/null
+          npm install
+        )
+        #      working-directory: "${{ inputs.srcdir }}"
```
assignUser, line 44:
```suggestion
```

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+    required: false
+    default: "presentations"
+  publish:
+    description: "Publish the site to the destination branch. If false, the site is built but not published."
+    required: false
+    default: "true"
+  output:
+    description: "AsciiDoc output directory"
+    required: false
+    default: 'output'
```
assignUser, line 23:
```suggestion
    default: "output"
```

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+  tempdir:
+    description: "Temporary Directory name"
+    required: false
+    default: 'output.tmp'
```
assignUser, line 27:
```suggestion
    default: "output.tmp"
```

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+  debug:
+    description: "AsciiDoc Debug mode"
+    required: false
+    default: 'false'
```
assignUser, line 31:
```suggestion
    default: "false"
```
For consistency

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+        # Remove all existing output so deletions will be captured
+        rm -rf "${{ inputs.output }}"
+        git rm --quiet -r --ignore-unmatch --cached "${{ inputs.output }}"/*
+        mv "${{ inputs.srcdir }}/${{ inputs.tempdir }}" "${{ inputs.output }}"
+        git diff # Show changes
+        git add "${{ inputs.output }}"
+        git status
+        if git commit -m "Commit build products"
+          then
+              git push
```
assignUser, line 98: I'm pretty sure this will fail as you are not setting up auth via `actions/checkout` or something.

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+    required: false
+    default: 'output.tmp'
+  debug:
+    description: "AsciiDoc Debug mode"
+    required: false
+    default: 'false'
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@v4
```
assignUser, line 35:
```suggestion
    - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
```
While it's an offical action and this doesn't require pinning in workflows I think it still makes sense within an action as the user has no control over it?

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+    - name: 
+      shell: bash
+      run: |
+        mkdir "${{ inputs.tempdir }}"
+        ln -s "${{ github.action_path }}/convert-slides.js" .
+        for file in `ls`; do
+          if [[ "$(echo $file | awk -F. '{print $NF}')" != "adoc" ]]; then
+            continue
+          fi
+          echo $file
```
assignUser, line 56: Does this not make the output very noisy? Maybe use groups to structure the log see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines

asciidoc/action.yml
```diff
@@ -0,0 +1,102 @@
...
+          node convert-slides.js $file
+          mv $(basename $file .adoc).html "${{ inputs.tempdir }}"/
```
assignUser, line 58:
```suggestion
          mv $(basename $file).html "${{ inputs.tempdir }}"/ 
```
Or did that serve a purpose there? I think it would just be dropped?

assignUser (commented):
Some comments, will have another look later

janhoy:
The name of this action is not precise. Asciidoc is a markdown format and can be used for writing books, READMEs or as in this case presentation slides with the reveal framework. So a better name for the action would probably be "presentations" or "slides" or "revealjs" or similar.

assignUser (commented):
Please also create a workflow to test the action with some sample input and some different options (obv publish is hard to test, we could create a test branch?).

You can also add npm to dependabot with asciidoc as a directory:
```yml
  - package-ecosystem: "npm"
    directory: "/asciidoc/"
    schedule:
      interval: "monthly"
```

dfoulks1:
@janhoy: Thanks for the suggestion! I'll update it to asciidoc-slides

asciidoc/action.yml
dfoulks1: would there be a way to test this / keep this hash up to date with dependabot or something? Trying to prevent having to touch this everytime there's an update.

asciidoc/action.yml
dfoulks1: This was all ripped from the pelican action. If there's a better / more GHA way to approach this, I'm 100% interested.

asciidoc/action.yml
dfoulks1: the basename command requires a suffix to strip.

asciidoc/action.yml
dfoulks1: this was pulled from the pelican action and has been tested against the infrastructure-presentations repo

dfoulks1 (commented):
responding to comments

asciidoc/action.yml
assignUser: yes dependabot can do it, just add the dir with action.yml to the `directories` list in dependabot.yml

asciidoc/action.yml
assignUser: ah that makes sense, if never needed that before :D

asciidoc/action.yml
assignUser: I assume the workflow using the pelican action called `actions/checkout` then.

You could make it a usage requirement to use this within an existing checkout but that should be documented specifically requiring that `persist-credentials:` is set to `true` otherwise this will fail.

asciidoc/action.yml
```diff
@@ -0,0 +1,115 @@
...
+    - run: npm install
+      shell: bash
+      working-directory: "${{ inputs.srcdir }}/${{ inputs.output }}"
+    - name: Setup Pages
+      uses: actions/configure-pages@v5
+    - name: Upload artifact
+      uses: actions/upload-pages-artifact@v3
+      with:
+        path: '${{ inputs.srcdir }}/${{ inputs.output }}'
+    - name: Deploy to GitHub Pages[
```
netomi, line 113: seems like a typo at the end '[' ?

asciidoc/action.yml
netomi: I would pin all actions also the ones coming from GitHub. The positive side effect is that your action does not have uncertainty wrt which exact action is being used. We have seen in the past that sometimes bugs are introduces in newly released versions and when you use just @v4 you will always get the latest released version of the v4 branch of that action. Having it pinned to an exact version makes actions more stable imho.

netomi: btw. you can use https://github.com/eclipse-csi/octopin for pinning.

install with `pipx install octocpin`

potiuk:

dave2wave:
Let's not close PRs from infrastructure staff.

potiuk:
Sure. I would love ot hear from @dfoulks1 if he is going to work on it, otherwise, it makes ittle sense to keep it. @dfoulks1 ?

dave2wave:
> > @dfoulks1 if you wish to finish this action then I suggest that you create a branch in the `infrastructure-actions` repository and move this code there. You can then create a Draft PR from the branch. This will have the advantage that we can test this using a real site like Incubator that uses asciidoc.
> 
> I'm curently using it as is in the infrastructure-presntations repo to test. let me pick this back up and over the finish line.

Two weeks ago in this PR.

potiuk:
Missed it - sorry

potiuk:
@dfoulks1 gentle nudge — back in April you mentioned picking this back up and taking it over the finish line. Still keen to land it? Moving the code to a branch in `infrastructure-actions` (as dave suggested) would let us test against a real asciidoc site like Incubator. No pressure — just checking whether it's still on your radar or should sit until you have bandwidth.

potiuk:
@dfoulks1 Could you rebase this onto latest `main` when you have a moment? Trying to get the open PRs current. Thanks!

PR #79, opened 2024-12-14
Pin MarkupSafe version

timsaucer:
# Request for updating an existing GitHub Action

## Overview

This is a small change to the pelican github action to pin the version of MarkupSafe during the docker build stage.

## Reasoning

MarkupSafe removed the pip dependency soft_unicode in later versions. With the current specification of MarkupSafe in the docker definition, you may not get the required dependency. There is one issue report relating to this: https://github.com/pallets/markupsafe/issues/304

This has led to recent failures in our pipelines such as: https://github.com/apache/datafusion-site/actions/runs/12319494192/job/34386667171?pr=50

## Alternative

One alternative is to leave the MarkupSafe definition as is and also add in an explicit install of soft_unicode.

## Work Around

We are currently working around this issue by adding in an additional github action step to install the pinned version of MarkupSafe.

Relates to https://github.com/apache/datafusion-site/pull/51

### DIFF
--- a/pelican/Dockerfile
+++ b/pelican/Dockerfile
@@ -50,7 +50,7 @@ COPY requirements.txt .
 RUN pip install -r requirements.txt --no-deps
 
 # Could perhaps be added to requirements.txt but that would affect other uses
-RUN pip install 'MarkupSafe<2.1.0' # needed for Pelican 4.5.4
+RUN pip install 'MarkupSafe==2.0.1' # needed for Pelican 4.5.4
 
 # Now add the local code; do this last to avoid unnecessary rebuilds
 COPY plugin_paths.py .


pelican/Dockerfile
```diff
@@ -50,7 +50,7 @@ COPY requirements.txt .
 RUN pip install -r requirements.txt --no-deps
 
 # Could perhaps be added to requirements.txt but that would affect other uses
-RUN pip install 'MarkupSafe<2.1.0' # needed for Pelican 4.5.4
+RUN pip install 'MarkupSafe==2.0.1' # needed for Pelican 4.5.4
```
potiuk, line 53: Idea: Maybe a good idea would be to add comment here to link to the issue/problem to solve it. Usually when I am doing such exclusion I create an issue that describes the "permanent" solution even if for now we want to temporarily limit it. Pinning a dependency inside dockerfile is quite problematic because almost for sure we will forget to unpin it or migrate (and we have now dependabot added to the repo to bump dependencies automatically, so likely some solution where dependabot will be able to update pinning automatically would be better.

assignUser: 100% agree

PR #82, opened 2024-12-31
Switch to v4 branch for upload-artifact action

potiuk:
We do not need to pin the action to specific commit, because it is an official github-maintained actions. Switching to branch will allow us to use latest bug-fix/feature version.

### DIFF
--- a/stash/save/action.yml
+++ b/stash/save/action.yml
@@ -115,7 +115,7 @@ runs:
 
     - name: Upload Stash
       id: upload
-      uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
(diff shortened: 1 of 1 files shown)


potiuk:
cc: @assignUser @dfoulks1 @sebbASF

assignUser:
I started pinning the action because a bug was introduced in the action breaking it so it can go both ways ^^. We should setup dependabot (made #83 ) regardless.

I would err on the side of pinning to shas and using dependabot to keep things current but that also causes issues with the current lack of tags (i.e. you'd need to update the airflow pins to get the new version...). This is probably a sign that we should talk about versioning again.

assignUser (approved):
+1 on this while we don't have proper versioning via tags.

gopidesupavan (approved):

PR #84, opened 2024-12-31
Add expanduser to the path passed to stash restore action

potiuk:
Often when you want to upload cache, you are doing it from the ~/.cache or similar directory. And you want to pass "~" rather than path to the home directory. The original upload-artifact and download-artifact are expanding the `~` to home dir, and since "stash/save" uses direction "upload-artifact" it also accepts the `~` prefix. However the "stash/restore" action uses gh download and passes the path via environment variable, so bash will not expand it. We need to manually expand user in the mung step


potiuk:
cc: @assignUser @dfoulks1 @sebbASF

assignUser (approved):
LGTM

gopidesupavan (approved):
good find :)

PR #86, opened 2024-12-31
Adds optional "overwrite" parameter to input of stash/restore action

potiuk:
Sometimes you want to restore stash (for example "~/.cache/" when it already exists - in which case `gh download` will fail with error extracting FILE: file exists

While you can delete the directory before restore action is called, it's a bit more convenient to add optional "overwrite" input - when set to "true" and target directory exists, it will delete it and subsequently re-create the directory.

## Overview

**Name of action:** 

**URL of action:**

**Version to pin to (branch, tag, hash):**

## Checklist


potiuk:
cc: @assignUser @sebbASF @dfoulks1 -> should help with handling cases similar to https://github.com/apache/airflow/actions/runs/12559549039/job/35015813646?pr=45289 in a convenient way

assignUser (commented):
I think `overwrite` implies that you could restore a cache into a dir and only the overlapping files would be overwritten with the new files. 
Calling it `clean` (actions/checkout uses that too iirc) might communicate better what happens? ?

potiuk:
> I think `overwrite` implies that you could restore a cache into a dir and only the overlapping files would be overwritten with the new files. Calling it `clean` (actions/checkout uses that too iirc) might communicate better what happens?

Yep. Why not

potiuk:
Done

assignUser (approved):
Thanks!

gopidesupavan (approved):

PR #87, opened 2025-01-02
Update approved_patterns.yml

gmcdonald:
Add 13 actions missing from the list that are in the internal configuration. Alpha-sort as per the internal whitelist


gmcdonald:
Do I need to remove the comma from each line ? When the list is synced to the internal whitelist proper then the comma must exist, so is some kind of post-processing happening to put the comma back to the end of each line ?

gmcdonald:
Right, the yaml doesnt need commas, but the receiving configuration within the Github Org settings does require commas. In fact my PR is a direct copy and paste from that Config so that I know it is correct how is should be sent back (via the API).

Stripping the commas here only to add them back again at insert time seems non-optimal. 

I agree this is not a true and valid yaml file, perhaps we just rename it to a .txt if we are not going to use yaml capabilities (?)

assignUser:
Not sure if this is already in prod (likely not?) but @dfoulks1 is expecting yaml in https://github.com/apache/infrastructure-gha-whitelist-manager

assignUser:
Closing as stale

PR #88, opened 2025-01-02
Add optional flag to only search for stash in current branch

potiuk:
In some cases, when you want to optimize building and using big artifacts, you want to build them in one job, upload as artifact and use them in the same workflow in other jobs, and you do not want to fallback to the target branch artifact. This is for example where you want to build docker image from current sources, and pull the image in other jobs of the same workflow.

If you make mistake in your workflow or miss uploading the artifact in this case, you will silently pull and use the "target" version of the artifact (so in this case it will be target sources of your code). This might lead to cases where your CI tests will not be testing your source code.

### DIFF
--- a/stash/restore/action.yml
+++ b/stash/restore/action.yml
@@ -22,8 +22,8 @@ inputs:
       functionality as there is no eviction (only expiry).
 
       The action checks the current branch for a stash, if there is no match,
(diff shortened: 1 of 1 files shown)


potiuk:
cc: @dfoulks1 @gopidesupavan @gopidesupavan

gopidesupavan (approved):
Nice fix :)

PR #91, opened 2025-01-13
Added explicit markupsafe version

dfoulks1:
The recent update to ubuntu-24 on the GHA runners seems to have caused an issue installing a newer version of markupsafe that doesn't have soft_unicode. `markupsafe==2.0.1` is now explicitly installed.

### DIFF
--- a/pelican/action.yml
+++ b/pelican/action.yml
@@ -45,7 +45,7 @@ runs:
       run: |
         (
           test "${{ inputs.debug }}" == 'true' || exec >/dev/null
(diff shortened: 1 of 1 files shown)


dave2wave (approved):
Other than my note LGTM

dave2wave:
> I do agree that overriding the system protection may well cause issues in the future

Yes, that is what I meant by "sustainable".

gstein:
I see no issues using --break-system-packages on git runners ... the runners are not multi-user, so whether a package is installed user-local or into the system ... it should not make a difference. Use the switch if needed to make things work.

Issue #94, opened 2025-01-14
Add CodeQL scanning


assignUser:
https://docs.github.com/en/code-security/code-scanning/enabling-code-scanning/configuring-default-setup-for-code-scanning

We should activate it for actions at the minimum. This is a repo setting, so maybe  also a candidate for the new asf yaml ^^
cc @dfoulks1

potiuk:
I think we can do it without configuring the repo. We do it for airflow https://github.com/apache/airflow/blob/main/.github/workflows/codeql-analysis.yml - I can PR it.

PR #101, opened 2025-01-14
Add CodeQL analysis

potiuk:
Fixes: #94

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,63 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


potiuk:
cc: @assignUser

assignUser:
Oh wow, I wasn't aware just adding the config enables it, great!

assignUser:
Ah I guess CodeQl disagrees @potiuk have you set codeql before? Does it choose a sensible default in your experience?

potiuk:
> Ah I guess [CodeQl disagrees](https://github.com/apache/infrastructure-actions/security/code-scanning/3) @potiuk have you set codeql before? Does it choose a sensible default in your experience?

I dismissed it as false-positive (the nice thing about those issues is that you can always dismiss them if you don't like them)- I personally prefer to choose the **matrix** of languages manually - and no - I have not tried default settings

assignUser:
The default check would pick up on new languages added to the repo but that is also part of the responsibility of the reviewers ^^

gopidesupavan:
> > Ah I guess [CodeQl disagrees](https://github.com/apache/infrastructure-actions/security/code-scanning/3) @potiuk have you set codeql before? Does it choose a sensible default in your experience?
> 
> I dismissed it as false-positive (the nice thing about those issues is that you can always dismiss them if you don't like them)- I personally prefer to choose the **matrix** of languages manually - and no - I have not tried default settings

I believe if we choose default matrix it runs for all languages. But also code-ql recommendation is to choose what languages that require to validate.
 https://github.com/github/codeql-action/blob/main/init/action.yml#L24.

I think we can update when we have other languages. ?

gopidesupavan (approved):

PR #103, opened 2025-02-04
Fix failing pelican test for PRs created by dependabot

netomi:
This fixes the workflow failures for PRs created from dependabot as such workflow runs will not be granted write permissions by default but the test actually writes to the branch.

Documentation about restrictions: https://docs.github.com/en/code-security/dependabot/troubleshooting-dependabot/troubleshooting-dependabot-on-github-actions#restrictions-when-dependabot-triggers-events

Blog post that dependabot triggered workflows will honor the requested permissions:

https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/

### DIFF
--- a/.github/workflows/pelican-action-test.yml
+++ b/.github/workflows/pelican-action-test.yml
@@ -6,6 +6,10 @@ on:
       - '.github/workflows/pelican-action-test.yml'
     # N.B. Cannot easily test pull_request: because the source may not have rights to create/update the test-site
   workflow_dispatch:
+
+permissions:
+  contents: write
+
 jobs:
   pelican-test:
     # do not run the pelican test if the PR was triggered / created by dependabot as the workflow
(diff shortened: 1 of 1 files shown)


netomi:
Actually the workflow should also be skipped for PRs from forks as then the workflow will also not get write tokens and fail.

jsoref:
The workflow should be fixed to have:
```yaml
permisions:
  contents: write
```

otherwise it will fail when people try running it in a fork with restrictive permissions (the fact that this repository/org/enterprise isn't using restrictive permissions is really worrisome).

netomi:
Ok I found this blog post where explicitly setting the permissions will be respected when dependabot triggers a workflow run:

https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/

so just adding the permissions snippet should solve the problem.

PR #106, opened 2025-02-13
Add release candidate action

stevedlawrence:
The Apache Daffodil project is taking steps to enable [Automated Release Signing](https://infra.apache.org/release-signing.html#automated-release-signing). For part of this we developed a GitHub action workflow that we plan to use for our different Daffdil projects. But we realized this is actually pretty generic and could possibly be useful for other ASF projects. It is a little opinionated on how things should be organized, but I think what it does isn't too controversial based on what I've seen of other ASF projects, at least for some of the smaller less complex ones.

So I thought I would document it better and see if this is something that would be of interest the the ASF community.

Also note that this action will not work as-is since the .js files need to be be compiled and those compiled files committed for GitHub actions to be able to actually use them. That adds a lot of noise to the PR so I left that off.

Thoughts?

### DIFF
--- a/release-candidate/README.md
+++ b/release-candidate/README.md
@@ -1,0 +1,147 @@
+# ASF Infrastructure Release Candidate Action
+
+This is a GitHub Action that can be used to create release candidates. Note
+that it is somewhat opinionated on how release candidates are organized. This
+is not intended to be used by all projects.
+
+## Prerequisites
+
+* Apache Security Team has approved your project for
+  [Automated Release Signing](https://infra.apache.org/release-signing.html#automated-release-signing)
+  and INFRA has set secrets for the repository, including a GPG signing key,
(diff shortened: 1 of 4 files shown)


release-candidate/README.md
```diff
@@ -0,0 +1,147 @@
...
+is not intended to be used by all projects.
+
+## Prerequisites
+
+* Apache Security Team has approved your project for
+  [Automated Release Signing](https://infra.apache.org/release-signing.html#automated-release-signing)
+  and INFRA has set secrets for the repository, including a GPG signing key,
+  SVN username/password, and nexus username/password.
+* The `runs-on` workflow setting should be Linux based (e.g. `ubuntu-latest`)
+* The repository must be checked out using `github/checkout` action prior to
```
netomi, line 14: I would say actions/checkout

PR #121, opened 2025-03-24
Group all dependabot updates together

netomi:
@assignUser @dfoulks1 

Considering that now a lot of dependabot updates will happen on a regular basis, it would make sense to group them together to avoid getting too many notifications and having to merge them manually. WDYT?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,3 +25,7 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    groups:
+      github-actions:
+        patterns:
+          - "*"


gopidesupavan:
> Yes I think that makes sense, the only downside I can see is that we might not want to update a version but I am sure you can tell dependabot to ignore one of the batched versions too in the same way you can in single dep PRs?

Which version your referring ?

netomi:
so this should be possible see the list of options here: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#managing-dependabot-pull-requests-for-grouped-updates-with-comment-commands

PR #142, opened 2025-04-02
feat: add golangci/golangci-lint-action

nodece:

## Overview

**Name of action: golangci/golangci-lint-action

**URL of action:**
https://github.com/golangci/golangci-lint
https://github.com/marketplace/actions/golangci-lint

* 1481404843c368bc19ca9406f87d6e0fc97bdcfd -> v7.0.0
* 55c2c1448f86e01eaae002a5a3a9624417608d84 ->v6.5.2

## Permissions

None.

## Related Actions

None.

## Checklist


assignUser (approved):
Just because this has caused confusion before, enabling these SHAs will NOT allow the matching tags (i.e. `@v7`) to work as the regex check does not resolve tags (even though I think it should).

assignUser:
That's not possible for us to do as all the workflow handling happens on GitHub's side before anything even runs.

assignUser:
(also the update to the allowlist may take some time to go through, but I will check with infra in this)

PR #148, opened 2025-04-24
Add CMake v2 action

git-hulk:

## Overview

**Name of action:** actions-setup-cmake

**URL of action:** https://github.com/jwlawson/actions-setup-cmake

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -468,6 +468,9 @@ jwgmeligmeyling/spotbugs-github-action:
   '*':
     expires_at: 2025-08-01
     keep: true
+jwlawson/actions-setup-cmake:
+  802fa1a2c4e212495c05bf94dba2704a92a472be:
+    expires_at: 2050-08-01
 kiegroup/github-action-build-chain:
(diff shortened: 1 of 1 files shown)


assignUser:
Thanks for the submission but I am rejecting this as the action is really not necessary. 

You can just use `pipx install --force cmake==3.31` as pipx is preinstalled on the runners, see [this job](https://github.com/assignUser/test-repo-a/actions/runs/14643217750https://github.com/assignUser/test-repo-a/actions/runs/14643217750) for a matrix using all the runners in your workflow. 

Even if pipx wasn't installed there are still ways with less baggage  than this action (which is a java script based actions with all the node_modules dependency issues that implies) e.g. something like [this script](https://github.com/apache/arrow/blob/520ae44272d491bbb52eb3c9b84864ed7088f11a/ci/scripts/install_cmake.sh#L60).

git-hulk:
@assignUser Thank you! I will give it a shot.

PR #153, opened 2025-05-19
add sha versions for more actions used by Pekko

pjfanning:
These actions are used by Apache Pekko and I'm looking to switch from versions to SHAs.

With the VirtusLab action, we don't use the old SHA any more so I replaced one with the latest one. I did a code search on Apache org GitHub and no one else uses this action.

Updated actions:
- coursier/cache-action
- coursier/setup-action
- jrouly/scalafmt-native-action
- VirtusLab/scala-cli-setup:

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -94,7 +94,7 @@ TobKed/label-when-approved-action:
     expires_at: 2025-08-01
     keep: true
 VirtusLab/scala-cli-setup:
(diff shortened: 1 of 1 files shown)


assignUser (approved):
Thanks for being pro-active with the switch to SHAs!

I have checked the SHAs and they are the latest release of each action. :+1:

PR #155, opened 2025-05-20
remove marcospereira/action-surefire-report

pjfanning:
Only Apache Pekko was using it. marcospereira/action-surefire-report is a clone of scacap/action-surefire-report and Pekko now uses that.


assignUser (approved):
Really appreciate your efforts!

PR #156, opened 2025-05-21
add mirromutth/mysql-action

slievrly:

## Overview

add mirromutth/mysql-action with set up a MySQL container for UT.

**Name of action:** 

mirromutth/mysql-action

**URL of action:**

https://github.com/mirromutth/mysql-action
https://github.com/marketplace/actions/setup-mysql

de1fba8b3f90ce8db80f663a7043be3cf3231248

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -801,3 +801,9 @@ golangci/golangci-lint-action:
   55c2c1448f86e01eaae002a5a3a9624417608d84:
     expires_at: 2050-08-01
     keep: true
+mirromutth/mysql-action:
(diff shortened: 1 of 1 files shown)


assignUser:
This action is clearly unmaintained, the last release was 2019! We won't be adding this to the list, sorry.

 I would recommend using [service containers](https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/about-service-containers) for this. Here is a relevant example setting up a [pg service](https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/creating-postgresql-service-containers).

PR #167, opened 2025-06-02
chore: Increase dependabot frequency

raboof:
As we'd ideally want infrastructure-actions to pick up upgrades before they're proposed in projects


assignUser:
This will flood us with PRs, I want to prune unused and archvied actions from the list first and improve some things. I'll work on that now, after we should be able to rapidfire merge prs

PR #168, opened 2025-06-06
Run tests for PRs

raboof:
There was a specific set of tests for the 'stash' action, which might still cover specific scenario's, but I think it would be good to do a basic run of 'all' tests as well - right now the gateway tests were not ran (and needed a fix)


raboof:
Some failures seem to be actual test failures.

Some failures seem to stem from the fact that the test requires access to a GH_TOKEN. I wonder if we could make these tests independent of GitHub, or make it easy to skip the tests that require such a token.

raboof:
I think this was already merged along with #169

PR #170, opened 2025-06-07
ci: trigger workflow on changes to gateway files

assignUser:
The workflows are meant to run on PRs but we haven't triggered them on changes to the actual scripts so far.

That's why we missed some changes in #169


assignUser:
@raboof something is still broken due to ruyaml changes, I'll have a look over the weekend but feel free to jump in ^^

PR #197, opened 2025-06-16
gateway: add mlugg/setup-zig

assignUser:

## Overview
This is the replacement for https://github.com/korandoru/setup-zig?tab=readme-ov-file#archived
which was archived. `mlugg` is a zig core developer so this is an almost 'official' action.

**Name of action:**
mlugg/setup-zig 

**URL of action:**
https://github.com/mlugg/setup-zig

v2.0.1 7dccf5e6d09267c55f815f2db29495f30ba2ebca

## Permissions
None

## Related Actions
See above, korandoru/setup-zig

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -882,3 +882,7 @@ golangci/golangci-lint-action:
   55c2c1448f86e01eaae002a5a3a9624417608d84:
     expires_at: 2050-08-01
     keep: true
+mlugg/setup-zig:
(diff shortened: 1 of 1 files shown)


raboof (approved):
addition seems OK

perhaps also expire korandoru/setup-zig in this PR?

(I wonder if we should sort this file by alphabet, to avoid merge conflicts at the end of the file)

PR #212, opened 2025-06-23
docs: how to find out if/where an action is still used

raboof:
Thanks to @assignUser for the recommendation


assignUser (approved):
Thanks!

PR #216, opened 2025-07-06
Update README.md for the stash action

dsahlberg-apache-org:
Add the apache/ prefix in uses so GHA find the action in the apache organisation. Change the used ref to @main, since there is no @v1 branch.

It took me, as a new user, a couple of tries to get it right. This changes should hopefully make it easier for new users.

### DIFF
--- a/stash/README.md
+++ b/stash/README.md
@@ -45,7 +45,7 @@ To restore a stash before your build process, use the `infrastructure-actions/st
 ```yaml
 steps:
 - uses: actions/checkout@v2
-- uses: infrastructure-actions/stash/restore@v1
+- uses: apache/infrastructure-actions/stash/restore@main
   with:
     key: 'cache-key'
     path: 'path/to/cache'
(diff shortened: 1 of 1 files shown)


stash/README.md
```diff
@@ -45,7 +45,7 @@ To restore a stash before your build process, use the `infrastructure-actions/st
 ```yaml
 steps:
 - uses: actions/checkout@v2
-- uses: infrastructure-actions/stash/restore@v1
+- uses: apache/infrastructure-actions/stash/restore@main
```
potiuk, line 48: Might be a good idea to mention a commit hash - this is really the only way we **should** be using actions (and `zizmor` action and "CodeQL` security check will flag both tag an branch usage.

I think it would be better to follow this in the example.

potiuk, line 48:
```suggestion
- uses: apache/infrastructure-actions/stash/restore@e88c4936fdeeb13c0910747af0bd07d992a81f9b
- ```

assignUser (approved):
Sorry for the inconvenience and thanks for the PR! Hope the action works well for you!

PR #221, opened 2025-07-22
chore: update erisu/license-checker-action

erisu:
# Request for updating an existing GitHub Action to the allow list

## Overview

This GitHub Action is being used to audit the licenses of the project's NPM dependency tree. The purpose of this action is to provide early detection during commits and PR to ensure that updated or newly added npm dependencies follows the ASF 3rd Party Category A License requirements.

**Name of action:** 

`erisu/license-checker-action`

**URL of action:**

https://github.com/erisu/license-checker-action

`1c222d0c2f5898a4c40b8bd6fd6888650bd6f68a` -> v2.0.0

## Permissions

     
- None 
     
This action only runs against the `github.workspace` which the user would have checkout with `actions/checkout`. There is no write access to the repo or read access of credentials.

## Related Actions

     
None

## Checklist

  - The CI is enabled to run CodeQL against the JS code.
  - Dependabot should also be enabled to monitor the npm dependencies.
  - Unit testing is not added but something I could easily prepare.

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -384,9 +384,9 @@ erisu/apache-rat-action:
     expires_at: 2025-08-01
     keep: true
 erisu/license-checker-action:
-  '*':
-    expires_at: 2025-08-01
-    keep: true
+  1c222d0c2f5898a4c40b8bd6fd6888650bd6f68a:
+    expires_at: 2050-01-01
+    tag: v2.0.0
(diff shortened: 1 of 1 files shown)


assignUser (approved):
I checked that any use of the action already uses the new sha, so the removal of the wild card is valid, thanks!

PR #223, opened 2025-07-26
Allow mlugg/setup-zig for all the versions

tisonkun:

## Overview

mlugg/setup-zig is the de-facto official action to set up Zig. The owner is Zig Software Foundation's member.

**Name of action:** mlugg/setup-zig

**URL of action:** https://github.com/mlugg/setup-zig

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):** I use `*` follow `setup-php` / `setup-dlang` and so on.

## Permissions

Default

## Related Actions

No

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -890,6 +890,6 @@ golangci/golangci-lint-action:
     expires_at: 2050-08-01
     keep: true
 mlugg/setup-zig:
-  7dccf5e6d09267c55f815f2db29495f30ba2ebca:
-    expires_at: 2050-01-01
-    tag: v2.0.1
+  '*':
+    expires_at: 2025-08-01
+    keep: true


assignUser:
Sorry we will not accept new wild card entries. The existing ones will be removed soon, as announced by infra a few months ago.

New versions of the action should automatically be picked up by dependabot. Though we might be a bit lagging on approving those atm, so if you need an new version urgently feel free to open a PR for it explicitly.

tisonkun:
@assignUser Since the approved commit changes over time, does it mean that ASF projects should update the hash over time?

See https://github.com/apache/opendal/blob/8de30cc764c3724db33907ca17e02648530061ed/.github/workflows/ci_bindings_zig.yml#L48

We pin a commit hash but one day it gets unapproved.

assignUser:
@tison We had a limit of 5 dependabot PRs which hid the update to this action, sorry. The latest release was added in https://github.com/apache/infrastructure-actions/pull/233 and can now be used, the old version will also work for a while but will be removed automatically in October.

tisonkun:
Thanks for your update. That makes sense :D

Issue #253, opened 2025-08-02
Add special case for `docker://` type actions to use `:` instead of `@` as version separator


assignUser:
As I noticed in #251 the current format leads to failed pulls:
```
/usr/bin/docker pull pandoc/core@sha256-48e15e83db0df6fb39b24adb0210ecbde85003a3a8139d526e29c98f95ac0a93
  invalid reference format
  Warning: Docker pull failed with exit code 1, back off 4.123 seconds before retry.
```

This also means we can purge all existing docker actions as no one is using them ^^

PR #254, opened 2025-08-02
fix(gateway): Use `:` as a reference separator for docker actions

assignUser:
Resolves: #253


potiuk:
> Oh no... why... the syntax is different for pulling by digest vs tag... tag uses : sha uses @.


Been there, done that... It's AWFUL yeah.

PR #272, opened 2025-08-25
Add new action posit-dev/setup-air v1.0.0

thisisnic:

## Overview

Automatic formatting of R Code.  

**Name of action:** 

posit-dev/setup-air

**URL of action:**

https://github.com/posit-dev/setup-air
https://posit-dev.github.io/air/integration-github-actions.html

63e80dedb6d275c94a3841e15e5ff8691e1ab237

## Permissions

Mostly read access though write access for PRs would be handy for automatically fixing issues if we want to use that.

## Related Actions

There are other actions for R styling within the currently allowed `r-lib/actions/*`, but this is a newer one which is significantly faster, gaining traction in R, released by Posit, and will cut down CI run times.

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -883,3 +883,7 @@ mlugg/setup-zig:
   8d6198c65fb0feaa111df26e6b467fea8345e46f:
     expires_at: 2050-01-01
     tag: v2.0.5
+posit-dev/setup-air:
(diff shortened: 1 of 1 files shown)


assignUser:
How disapointing that air is not on crates.io, that makes using it in a hook much harder, will require a separate repo vs just installing it ... :(

PR #274, opened 2025-08-27
Add freebsd-vm action into allowlist

PragmaTwice:

## Overview

**Name of action:** vmactions/freebsd-vm

**URL of action:** https://github.com/vmactions/freebsd-vm

`05856381fab64eeee9b038a0818f6cec649ca17a`

https://github.com/vmactions/freebsd-vm/commit/05856381fab64eeee9b038a0818f6cec649ca17a

## Permissions

None.

## Related Actions

None.

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -870,3 +870,7 @@ posit-dev/setup-air:
   63e80dedb6d275c94a3841e15e5ff8691e1ab237:
     expires_at: 2050-01-01
     tag: v1.0.0
+vmactions/freebsd-vm:
+  05856381fab64eeee9b038a0818f6cec649ca17a:
+    expires_at: 2050-01-01
+    tag: v1.2.3


assignUser (commented):
The tag matches. 

My problem with this Action is that it appears like the entire 'vmactions' org is maintained by a single person with some additional contributions in select repos. That doesn't really fit our criteria. But I also see that this action is without any real alternatives and actively maintained.

@PragmaTwice I will not allow list this action with the workflow in the current state. Minimize permissions of the GITHUB_TOKEN of the job using: (this should be the default for ALL jobs)
```yaml
permissions:
  contents: read
```

PragmaTwice:
> I will not allow list this action with the workflow in the current state. Minimize permissions of the GITHUB_TOKEN of the job using: (this should be the default for ALL jobs)

Done in https://github.com/apache/kvrocks/pull/3141/commits/a5ca0280ec923cd0de3e9925667be1a1fcccd52a.

PR #281, opened 2025-08-30
gateway: Remove cpp-linter-action v2.16.0

assignUser:
This version added a new sub-action that is not allow-listed, see https://github.com/cpp-linter/cpp-linter-action/pull/306#issuecomment-3235602010

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -300,11 +300,8 @@ cpp-linter/cpp-linter-action:
     expires_at: 2025-11-07
     tag: v2.15.0
   0f6d1b8d7e38b584cbee606eb23d850c217d54f8:
-    expires_at: 2025-11-17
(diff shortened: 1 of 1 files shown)


2bndy5:
Instead of using cpp-linter-action, you can brew your own workflow that uses [the cpp-linter python package](https://pypi.org/project/cpp-linter/) directly (designed as a CLI tool). See our [test repo's workflow](https://github.com/cpp-linter/test-cpp-linter-action/blob/master/.github/workflows/cpp-lint-package.yml) as an example.

Sorry for any inconvenience.

assignUser:
@2bndy5 I appreciate you taking time to about an alternative here! 

Due to unrelated reasons I am no longer watching/maintaing this repo. Someone else will pick this up.

PR #312, opened 2025-09-29
Add `tj-actions/changed-files`

blaginin:

## Overview

I want to add this action, because for Datafusion we [need a way](https://github.com/apache/datafusion/issues/6880) to check if the PR has touched some files, and Merge Queue [doesn't work](https://github.com/orgs/community/discussions/45899) with `paths-ignore`

**Name of action:**  tj-actions/changed-files

**URL of action:** https://github.com/tj-actions/changed-files

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):** [`24d32ff`](https://github.com/tj-actions/changed-files/commit/24d32ffd492484c1d75e0c0b894501ddb9d30d62)

## Permissions

Action can work with read only access 

## Related Actions

There's a similar one https://github.com/Ana06/get-changed-files already in the list, but I believe it's not maintained 

## Checklist

- [x] The action is listed in the GitHub Actions Marketplace: https://github.com/marketplace/actions/changed-files
- [X] The action has a clearly defined license: MIT
- [x] The action is actively developed or maintained: commits every months since 2021
- [x] The action has CI/unit tests configured: https://github.com/tj-actions/changed-files/tree/main/test

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -856,6 +856,10 @@ testlens-app/setup-testlens:
 timonvs/pr-labeler-action:
   '*':
     keep: true
+tj-actions/changed-files:
+  24d32ffd492484c1d75e0c0b894501ddb9d30d62:
+    expires_at: 2050-01-01
+    tag: v47.0.0
 TobKed/label-when-approved-action:
   '*':
     keep: true
@@ -874,6 +878,13 @@ vimtor/action-zip:
 vishalsinha21/dynamic-checklist:
   '*':
     keep: true
+vmactions/freebsd-vm:
+  05856381fab64eeee9b038a0818f6cec649ca17a:
+    expires_at: 2025-12-22
+    tag: v1.2.3
+  487ce35b96fae3e60d45b521735f5aa436ecfade:
+    expires_at: 2050-01-01
+    tag: v1.2.4
 wei/curl:
   '*':
     keep: true


blaginin:
Yes! I pinned a patched version though.

I was also wondering if this should be a reason not to use it - but to me, it feels like a good sign of community engagement and repo maintenance

snazy:
I'm just curious how things work ;)
I think, the double-for-loop can be replaced with this, which checks tests all changed files against a list of glob expressions.
```
shopt extglob
if [[ ${changed_files[@]} == @(*.md|docs/*|.github/ISSUE_TEMPLATE/*|.github/pull_request_template.md) ]]; then
  echo yes
else
  echo no
fi
```
But yea, without an action, it's bash-magic.

blaginin:
yes, sure! would you be ok with merging this pr or do you want me to use bash? i d prefer PR but dont mind writing bash hehe 😊

snazy:
I can't (not an infra guy) :)

raboof:
if you're comfortable vouching for this action then an approval is welcome even if you're no committer on this repo

snazy (approved):
I gave it a try.
I'm not a typescript guru, but I couldn't find anything in the action's code base that looks suspicious - like "interesting" `exec` or GH API calls or the like.
The workflows they have look good to me, nothing obvious that looks like a way to inject malicious code into the build or the like. But I haven't looked into the actions they depend on.
Tests look reasonable, with a reasonable test (case) coverage.
The project itself has quite a variety of contributors and seems overall quite mature.
TL;DR haven't found any "red flags"

blaginin:
> In other words: a false-positive warning is emitted when using the parameter

hmm i m pretty sure it doesn't work without the input properly added: https://github.com/apache/datafusion-sandbox/commit/d04372a5b9dca0a38d4a5f7485f5df4ab2427b8d#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR58

blaginin:
> Ok; my experience was that it worked correctly for a use case to skip GitHub Actions checks on commits that only modify documentation: https://github.com/openculinary/openfoodfacts-server/pull/6

Yeah, that's a fair point. But even in that repo you've referenced, the team switched to `changed-files` because 
> dorny/path-filter action, which turned out to be buggy

https://github.com/openfoodfacts/openfoodfacts-server/pull/11630

blaginin:
Hi! Thanks for the rebase!!@jayaddison's suggestion has helped and I managed  to use dorny. Happy to close this now 🙏

Issue #252, opened 2025-08-02
Having an expiry date and keep seems counter-intuitive


gmcdonald:
Some actions have an expiry date and also the `keep:true` flag set.
It should be one or the other.

Many actions expired yesterday but are kept due to the keep flag, these need going through manually

PR #320, opened 2025-10-06
the `apache/*` `*` rule should not expire

raboof:
towards #252


dfoulks1:
does the rule need to be there at all, or is allowing your own actions not an implicit thing?

PR #339, opened 2025-10-26
Allow use of `ppkarwasz/fetch-metadata` alongside `dependabot/fetch-metadata`

ppkarwasz:

## Overview

This PR allows the usage of the `ppkarwasz/fetch-metadata` GitHub Action as an alternative to `dependabot/fetch-metadata` in ASF repositories.

The `ppkarwasz/fetch-metadata` action is a personal improvement of the original `dependabot/fetch-metadata`, adding support for grouped Dependabot pull requests, a feature that is currently missing from the upstream action. The implementation has already been reviewed and approved by the Dependabot team (see dependabot/fetch-metadata#632), but the upstream project has been inactive for several months, likely due to reduced maintenance capacity at GitHub. This has prevented the improvement from being merged and released.

**Name of action:** `ppkarwasz/fetch-metadata`

**URL of action:** https://github.com/ppkarwasz/fetch-metadata

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):** `14da1d746fb4c6f05dbc353e4fd619a1065c8ff6`

## Permissions

The action only requires `read` permissions, so no explicit permission besides what is publicly available.

## Related Actions

`dependabot/fetch-metadata`

## Why this change is needed

In Apache Logging Services, every pull request must include a changelog entry. Previously, under CTR, we used a workflow that automatically added the changelog entry and merged the PR.

Since switching to RTC, this automation can no longer complete the merge step, resulting in repositories accumulating unmerged Dependabot PRs that must be:

* manually reviewed,
* updated with an empty commit to re-trigger required status checks,
* and merged by hand.

We already have an improved workflow in place (see apache/logging-parent#419) that provides:

* **Security enhancements** through separation of privileged and unprivileged workflows (`ppkarwasz/fetch-metadata` is used only in the unprivileged workflow),
* **Automatic merge using `auto-merge` instead of manual merging**, and
* **Support for grouped Dependabot PRs** (reducing noise to ~1 PR per repository per month).

The final item, grouped PR support, requires the `ppkarwasz/fetch-metadata` action.

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -623,6 +623,10 @@ potiuk/cancel-workflow-runs:
   '*':
     expires_at: 2025-08-01
     keep: true
+ppkarwasz/fetch-metadata:
+  14da1d746fb4c6f05dbc353e4fd619a1065c8ff6:
+    expires_at: 2025-10-01
+    tag: v0.1.0
 pre-commit/action:
   2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd:
     expires_at: 2050-01-01


ppkarwasz:
After revisiting `dependabot/fetch-metadata`, it appears that while the action still does **not** correctly resolve the **previous** version of a dependency, it does correctly determine the **new** version. That should be sufficient for generating the changelog entries used in Logging Services.

If I can confirm this behavior, I’ll close this PR and open a new one proposing the use of the upstream `dependabot/fetch-metadata` action instead, as it remains the officially maintained option.

ppkarwasz:
Hi @dfoulks1,

I tried to review `dependabot/fetch-metadata` from a security point of view and I’m leaning toward **removing both** `ppkarwasz/fetch-metadata` and `dependabot/fetch-metadata` from this proposal, for a few reasons:

* I’m not a TypeScript expert, and I’d prefer not to maintain or vouch for a complex dependency chain I don’t fully understand.
* Both actions currently include **known vulnerabilities** in their dependency stack (CVE-2025-25288, CVE-2025-25289, CVE-2025-25290).
* While these vulnerabilities are likely **not directly exploitable** in our usage context, I don’t have the bandwidth to confirm that.
* Fixing them would require **major version bumps** across the dependency tree, which could introduce breaking changes.
* The issues appear to be **transitive** dependencies of [`@actions/github`](https://github.com/actions/toolkit), but I haven’t found any clear information on their practical impact.

Given that, I’m not comfortable declaring either action “safe” for ASF-wide use, especially since `dependabot/fetch-metadata` appears dormant since June.

@vy, what do you think about us maintaining a **minimal TypeScript action** ourselves?
It could have far fewer dependencies and focus solely on the small subset of metadata we actually use for changelog generation.

vy:
> @vy, what do you think about us maintaining a **minimal TypeScript action** ourselves? It could have far fewer dependencies and focus solely on the small subset of metadata we actually use for changelog generation.

@ppkarwasz, until the time we have an **active** maintainer in the [Logging Services] PMC who can and is willing to maintain a {Java,Type}Script stack, **and** it is objectively clear that keeping a certain component in {Java,Type}Script is simpler & more sustainable compared to alternatives, I am inclined to [KISS](https://en.wikipedia.org/wiki/KISS_principle), and stick to tools that we know and use best: Java & shell scripting.

PR #354, opened 2025-11-07
Add `tag` values for `golangci/golangci-lint-action` in `actions.yml`

snazy:
See https://github.com/golangci/golangci-lint-action/releases


raboof (approved):
LGTM!

I don't think this would break anything, the version tag is mostly 'informational'

PR #356, opened 2025-11-07
Verify GH action tag/SHA combinations

### LINKED ISSUE #110: Verify SHA belongs to released version
> On top of that that makes it quite hard if the pinned hash actually corresponds to an actual released version of the action which could lead to the case where somebody might getting a commit hash approved that is not actually a released version. I would certainly add version comment for each action and add a validation to see if it really corresponds to the tag.

@netomi in slack

snazy:
This change introduces a new function `verify_actions` to validate the contents against GitHub.

TL;DR
The function verifies that the SHAs specified in `actions.yml` exist in the GH repo. Also ensures that the SHA exists on the Git tag, if the `tag` attribute is specified. The rest of the function is a lot of output and error(failure) and warning collection.

Although it issues quite a few GH API requests, the rate limiter should not kick in (with an authenticated GH token, GH workflows have a limit of 15k requests). I opted to rely on the HTTP/1.1 `urllib.request` stuff, which has no connection-reuse. The alternative would have been to add a dependency.

The algorithm roughly works like this, for each action specified in `actions.yml`:
* Issue a warning and stop, if the name is like `OWNER/*` ("wildcard" repository). Can't verify Git SHAs in this case.
* Issue a warning and stop, if the name is like `docker:*` (not implemented)
* Issue an error and stop, if the name doesn't start with an `OWNER/REPO` pattern.
* Each expired entry is just skipped
* If there is a wildcard reference and a SHA reference, issue an error.

Then, for each reference for an action:
* If no `tag` is specified, let GH resolve the commit SHA. Emit a warning to add the value of the `tag` attribute, if the SHA can be resolved. Otherwise, emit an error.
* If `tag` is specified:
  * Add the SHA to the set of requested-shas-by-tag
  * Call GH's "matching-refs" endpoint for the 'tag' value
    * Emit en error, if the object type is not a tag or commit.
    * Also resolve 'tag' object types to 'commit' object types.
    * Add each returned SHA to the set of valid-shas-by-tag.
* For each "requested tag" verify that the sets of valid and requested shas intersect. If not, emit an error.

Fixes #110


snazy:
@raboof do you think it's worth tackling this one?

raboof:
This looks helpful to me. Does it actually share code with `gateway.py` or could it be a separate file?

snazy:
> This looks helpful to me. Does it actually share code with `gateway.py` or could it be a separate file?

It does use the load_actions and a data class. But nothing that presents moving the code to a separate .py file.

raboof (commented):
This seems like a helpful check. I'm not sure it should be executed as part of the `Handle Dependabot update` workflow, perhaps it should be in its own periodically-scheduled workflow? we'd expect dependabot to specify the 'right' values, but they might change over time.

snazy:
Moving to draft, as this depends on #478, #479 and #480 now

potiuk:
I think it's a helpful check

snazy:
I've rebased this PR.

Two actions are a bit problematic:
* `dtolnay/rust-toolchain` references the `stable` branch, not a particular tag -  the Git commit exists though.
* `matlab-actions/run-tests`: the tag `v3.1.0` was added on Apr 12, 2026 via PR #695, but moved on Apr 14, 2026 to `64426f612825c99055c2b4887c08c135ab01be0b`/`a6c82b97e58e6b95bf49fcc8f5ea82370dcccd12`. The diff to the original SHA `353aee49b0edf62278c118a51b484d90bf6da1b7` is [here](https://github.com/matlab-actions/run-tests/compare/64426f612825c99055c2b4887c08c135ab01be0b...353aee49b0edf62278c118a51b484d90bf6da1b7). However, v3.1.0 is mentioned as a pre-release. 
image

potiuk:
> dtolnay/rust-toolchain references the stable branch, not a particular tag - the Git commit exists though.

I guess we should make it into 1.94.1 (different sha aad518f59d88bae90133242f9ddac7f8bbc5dddf)  - can you update it @snazy ?

This has been added here https://github.com/apache/infrastructure-actions/pull/620 -@kevinjqliu - are you ok with that? I guess you are the user of that action  - so you will have to update it in your repo? 

Maybe also worth  checking if others do not use it ?

> matlab-actions/run-tests: the tag v3.1.0 was added on Apr 12, 2026

This has been removed in main - they apparently changed "release" to "pre-release" and kept on moving it.

kevinjqliu:
> This has been added here https://github.com/apache/infrastructure-actions/pull/620 -@kevinjqliu - are you ok with that? I guess you are the user of that action - so you will have to update it in your repo?

looks like a bunch of repos are already using that commit hash https://grep.app/search?f.repo.pattern=apache%2F&q=dtolnay%2Frust-toolchain%4029eef336d9b2848a0b548edc03f92a220660cdb8

and some still using `@stable` https://grep.app/search?f.repo.pattern=apache%2F&q=dtolnay%2Frust-toolchain%40stable

potiuk:
@snazy gentle nudge on this — still think the tag/SHA verification check is worth having. `check_action_tags` is currently red; from the thread the open edge cases were `dtolnay/rust-toolchain` (tracks the `stable` branch, no tag) and the `matlab-actions/run-tests` pre-release tag move. Are you planning to push this over the line, or would it help to split out the parts that are ready (the `ignore_gh_api_errors` flag + core verification) so they can land incrementally?

snazy:
I've rebased the branch.

The changes to `actions.yml` now only contain the changes for the `scacap/action-surefire-report` action to ignore GH API errors, because that org has an "IP allowlist" that prevents us from using the GH API against that org.

I've made the `dtolnay/rust-toolchain` action case that references their `master` branch a warning, instead of a hard failure.

potiuk:
Thanks for the rebase and scoping this down, @snazy — the approach is solid and the test coverage on the happy paths is nice. I think the check is worth having. Two things I'd want sorted before approving, plus a few smaller items:

**🔴 Blockers**

1. **Leftover debug `raise` in `gateway/action_tags.py`** (the invalid-Git-SHA branch):
   ```python
   else:
       result.failure(f"... references an invalid Git SHA '{ref}'", "  ..")
       raise Exception("foo")
   ```
   On any invalid SHA this records the failure and then crashes the run with `Exception("foo")` instead of failing gracefully. The branch isn't covered by a test, so green CI doesn't catch it. The `result.failure(...)` above already does the right thing — the `raise` should go (a small regression test for this branch would be great too).

2. **The workflow won't trigger on the PRs it's meant to guard** (`check_action_tags.yml`). The path filters reference files that don't exist:
   ```yaml
   push:         paths: [".github/workflows/dummy.yml"]
   pull_request: paths: [".github/workflows/update_actions.yml", ".github/workflows/dummy.yml", "gateway/*"]
   ```
   There's no `dummy.yml` or `update_actions.yml` — the sync workflow is `update.yml`, and the inputs that should be verified are `actions.yml` and `.github/actions/for-dependabot-triggered-reviews/action.yml`. As written, a dependabot bump or an `actions.yml` edit won't run this check; only `gateway/*` changes and manual dispatch do. Adding `actions.yml` + the composite to the triggers would make it fire when it matters.

**🟡 Non-blocking, worth a look**

- `os.environ['GH_TOKEN']` (lines 88, 157) raises `KeyError` on local runs without the token — `.get()` would be friendlier.
- `today: date = date.today()` default arg is evaluated once at import, not per call — `today: date | None = None` then default inside is the usual fix.
- `run_action_tags.py` calls `update_actions`/`update_patterns`, so a "verify" run also rewrites `actions.yml`/`approved_patterns.yml` — slightly surprising side effect; might be worth a comment explaining it's intentional. (Also a small typo: "GH_TOKEN environment variable should be must.")
- The GHA step-summary writer emits a stray ` ``` ` fence when there are failures but no warnings.

Nothing here is structural — happy to approve once the two blockers are addressed.

snazy:
Thanks for the review! Pushed another commit to address the comments.

potiuk (approved):
Approving — this closes a real gap: nothing previously checked that the SHA↔tag pairs in `actions.yml` are genuine, and a forged or typo'd pair would have gone unnoticed. The design fits the gateway conventions well (reuses `ActionsYAML`/`update_actions`/`update_patterns`, extends `RefDetails`), the permissions are correctly read-only + token-gated, and test coverage is solid. Verified locally that it merges cleanly against `main` (no conflicts — just needs a rebase to clear the `mergeable: UNKNOWN`).

A few non-blocking nits I'd like addressed before merge:

1. **`pip install ruyaml` → `uv`** (`check_action_tags.yml`): the rest of the repo standardizes on `uv run` / `uvx` (see `pytest.yml`, `update.yml`); bare `pip` skips the `# /// script` dependency pinning. `uvx --with ruyaml` (as `run_action_tags.py`'s own docstring suggests) would match convention. There's also a trailing space on that line that pre-commit may flag.
2. **Rate-limit / transient-error resilience**: the check issues several sequential `api.github.com` calls per ref, and a single `403`/`429` currently hard-fails the whole run. With `GITHUB_TOKEN`'s 5000/hr it should fit, but a transient secondary-limit would red the check with no retry/backoff. Not blocking — just flagging that `ignore_gh_api_errors` is the only mitigation today.
3. **Minor**: the live-API tests in `test_action_tags.py` assert on specific upstream SHAs (e.g. `setup-uv` v7.1.2), which are brittle if upstream re-tags — worth a comment noting they may need updating.

Thanks for tackling this — it's a genuinely useful guardrail.

snazy:
The remaining Zizmor CI failure is unrelated to this change, created #918 to address those separately.

potiuk:
@snazy Could you rebase this onto latest `main` when you get a chance? It's approved and green, just want it current before merge. Thanks!

snazy:
@potiuk done

potiuk:
Re-reviewed after the rebase — still LGTM, merging. A few **non-blocking** follow-up nits for whenever (none worth holding this up):

1. `action_tags.py` parses API responses with `ruyaml.YAML().load()` — works since JSON ⊂ YAML, but `json.loads` (stdlib) is a touch more robust/faster for response bodies.
2. `git/matching-refs/tags/{tag}` is a *prefix* match, so a coarse tag like `v3` also pulls in SHAs from `v3.x` siblings — effectively exact for full semver tags, but it slightly loosens the SHA↔tag check for short tags.
3. No pagination on `matching-refs` — a tag prefix with >30 matching refs would be truncated (edge case).
4. A wildcard ref followed by an untagged SHA, in that order, can skip the "wildcard but also specific SHAs" warning (warning-only).
5. Tiny typo: `tab_object_type` → `tag_object_type`.

Thanks for pushing this through — the branch-vs-tag fallback is a nice touch.

PR #363, opened 2025-11-10
simplify 'no expiration'

raboof:
Instead of having 2050-01-01 as 'magic no expiration date', just leave this field out.


snazy (approved):
LGTM - thanks for the clarification!
+1

snazy (approved):

PR #388, opened 2025-11-24
deprecate `VirtusLab/scala-cli-setup`

raboof:
AFAICS this is not used anywhere (anymore?)


sebbASF:
it looks like the only references to VirtusLab are the ones in this repository ...

PR #452, opened 2026-01-14
Allow pypa/gh-action-pip-audit

tisonkun:

## Overview

This is currently used in https://github.com/apache/libcloud/tree/trunk/.github/actions and I think it's a pypa official action that we can trust.

**Name of action:** pypa/gh-action-pip-audit

**URL of action:** https://github.com/pypa/gh-action-pip-audit

Following the other pypa actions listed below, I choose:

```yaml
  release/v1*:
    expires_at: 2025-08-01
    keep: true
```

## Permissions

I think the read permission is enough.

## Related Actions

This should be like `pypa/gh-action-pypi-publish` but only for verifying, not publishing.

## Checklist

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -37,9 +37,9 @@ golangci/*:
     expires_at: 2025-08-01
     keep: true
 pypa/gh-action-pip-audit:
-  release/v1*:
-    expires_at: 2025-08-01
-    keep: true
+  1220774d901786e6f652ae159f7b6bc8fea6d266:
+    tag: v1.1.0
+    expires_at: 2026-03-31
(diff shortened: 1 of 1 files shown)


tisonkun:
@dfoulks1 Updated. Hopefully this is the correct format.

micafer:
This action is not available now as it has expired. Could you add it again?

tisonkun:
Seems 7367fb1 remove the action.

@dfoulks1 I remember I was told that the action would be automatically upgraded. How should I fix the gap?

If I have to do this sync manually from time to time, it looks cumbersome.

PR #481, opened 2026-02-12
setup-bazel: remove wildcard allowlisting

raboof:
incorrectly introduced in
https://github.com/apache/infrastructure-actions/pull/419/changes

Also adding 0.18.0 as used in
https://github.com/apache/skywalking-data-collect-protocol/blob/master/.github/workflows/ci.yaml


snazy (commented):
As there's only one usage (of 0.18.0), should the reference to 0.15.0 be removed?

PR #484, opened 2026-02-12
avoid having both 'keep' and 'expires_at'

raboof:
Remove 'expires_at' if there is also 'keep', since 'keep' currently takes precendence.

We'll want to remove the 'keep' entries for most of these cases eventually, but let's do that gradually to reduce impact.

See also https://github.com/apache/infrastructure-actions/issues/252

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -21,6 +21,9 @@ astral-sh/setup-uv:
     expires_at: 2026-05-15
   eac588ad8def6316056a12d4907a9d4d84ff7a3b:
     tag: v7.3.0
+    expires_at: 2026-05-24
+  5a095e7a2014a4212f075830d4f7277575a9d098:
+    tag: v7.3.1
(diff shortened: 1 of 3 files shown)


actions.yml
```diff
@@ -746,28 +635,22 @@ vimtor/action-zip:
...
 slackapi/slack-github-action:
   91efab103c0de0a537f72a35f6b8cda0ee76bf0a:
     tag: v2.1.1
 golangci/golangci-lint-action:
-  1481404843c368bc19ca9406f87d6e0fc97bdcfd:
-    expires_at: 2050-08-01
-    tag: v7.0.0
-    keep: true
   55c2c1448f86e01eaae002a5a3a9624417608d84:
     expires_at: 2050-08-01
```
snazy:
```suggestion
    keep: true
```
We probably don't care about this commit ID in 24 years ;)

PR #486, opened 2026-02-12
Workflow to check any ASF project's action usage

### LINKED ISSUE #482: Allow projects to verify their actions usage
When authoring GH workflows, people may add new or change action references (`uses:`).
Also dependabot and Renovate change action references.

If the workflows/jobs/steps are not run when such an action reference is added or changed, either by humans or bots, those are not validated against the set of allowed actions.

It would be nice for ASF projects to be able to validate their workflow files before a change is being merged, as part of their usual CI jobs.

I can imagine, that a workflow in the `apache/infrastructure-actions` could provide the necessary logic to check all `uses:` against the allowed  actions. That workflow could then be called, via `workflow_call`, from projects' workflows, so those do not need to copy any code.

snazy:
This change adds a GitHub workflow to be called from ASF projects' CI to verify that the referenced actions are approved by ASF Infrastructure.

Fixes #482

### DIFF
--- a/.github/workflows/check-project-actions.yml
+++ b/.github/workflows/check-project-actions.yml
@@ -1,0 +1,62 @@
+# Workflow to be called from ASF project repository workflows to check
+# whether the referenced GitHub actions are approved.
+#
+# The README.md of ASF Infrastructure Actions repository https://github.com/apache/infrastructure-actions
+# contains usage instructions.
+#
+# See: ASF Infrastructure GitHub Actions Policy: https://infra.apache.org/github-actions-policy.html
+
+
+name: check-project-actions.yml
+on:
+  workflow_call:
+    inputs:
+      repository:
+        required: false
+        description: The `repository` parameter for `actions/checkout`
+        type: string
+      ref:
+        required: false
+        description: The `ref` parameter for `actions/checkout`
+        type: string
+      fetch-depth:
+        required: false
+        description: The `fetch-depth` parameter for `actions/checkout`
+        type: number
+        default: 1
+      submodules:
+        required: false
+        description: The `submodules` parameter for `actions/checkout`
+        type: boolean
+        default: false
+
+jobs:
+  check-project-actions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout apache/infrastructure-actions"
+        uses: actions/checkout@v2
+        with:
+          repository: 'apache/infrastructure-actions'
+          ref: 'main'
+          path: infrastructure-actions
+      - name: "Checkout repository to be checked"
+        uses: actions/checkout@v2
+        with:
+          repository: '${{ inputs.repository }}'
+          ref: ${{ inputs.ref }}
+          fetch-depth: ${{ inputs.fetch-depth }}
+          submodules: ${{ inputs.submodules }}
+          path: repository
+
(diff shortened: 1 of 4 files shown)


raboof:
> I'm a committer in Apache Superset and it'd be great to integrate this workflow into our automated Dependabot PRs for bumping GHA version to ensure whether the version bump is allow-listed or not.

Ah that's a good point. Perhaps it could also link to a URL like https://github.com/apache/infrastructure-actions/pulls?q=is%3Apr+is%3Aopen+action-send-mail to make finding the update to review easier?

.github/workflows/check-project-actions.yml
```diff
@@ -0,0 +1,62 @@
+# Workflow to be called from ASF project repository workflows to check
```
raboof: since this workflow is meant to be re-used in other projects, perhaps it would be better to place it in its own top-level directory (next to `pelican`, `stash`, etc) instead of in the `.github/workflows` of this repo? Then it could also have its own README there.

README.md
```diff
@@ -11,6 +12,47 @@ This repository hosts GitHub Actions developed by the ASF community and approved
...
+You can let your CI workflows check if the Actions used in your project are approved for use in the ASF.
+
+Either create a new workflow in your project repository, e.g. `.github/workflows/check-project-actions.yml`,
+like the following example, or call the workflow from a job from your existing CI workflow in your repository.
+
+```yaml
+ name: Check action references
+ on:
+  workflow_dispatch:
+  push:
```
raboof: can we show how to lock down the permissions for this workflow? It shouldn't need any particular ones, right?

raboof (commented):
might be better to move it to its own top folder? otherwise: if there's demand for it, looks reasonable to have.

check-actions-usage/sample-ci-workflow.yml
```diff
@@ -0,0 +1,49 @@
...
+    branches:
+      - main
+    paths:
+      - ".github/**"
+  pull_request:
+    paths:
+      - ".github/**"
+
+permissions:
+  contents: read
```
raboof: 👍

raboof (approved):

snazy:
> requiring `pull-requests: write` would mean we'd have to audit this action and its dependencies much more closely for security issues. I think the isolation is well worth the few clicks...

Fair point, I've removed the PR comment stuff.

potiuk:
I really love that action. Once we have it - we can tell any project that has problems with their actions, that they should use it. Alternatively we could require all our PMCs to use that workflow - this should  be quite easily enforceable with github action checker script we already run @raboof @dfoulks1 @dave2wave

potiuk (approved):

dave2wave (dismissed):
@snazy would you please edit your PR to apply to the `apache:action-usage` branch. This will allow us to test the new action, iterate as a group on the fixes, and then create the final PR.

snazy:
> @snazy would you please edit your PR to apply to the `apache:action-usage` branch. This will allow us to test the new action, iterate as a group on the fixes, and then create the final PR.

@dave2wave Sorry, I totally missed your ping. Feel free to ping me on Slack, if I'm unresponsive.

@potiuk Thanks for merging!

PR #537, opened 2026-03-16
runs-on 2.1.0 (update)

blaginin:

## Overview

Add the update: https://github.com/runs-on/action/releases/tag/v2.1.0


potiuk (changes requested):
We cannot accept wildcards. 

Best way (See the README) @blaginin - is to setup dependabot (example in README) in your repo and make sure you are also updating the actions when the new action is released.

potiuk:
I don't think we will ever approve wilcard actions. Closing it - we can always re-open if needed.

PR #607, opened 2026-03-27
Revert dependabot cooldown configuration

kevinjqliu:
## Summary

Reverts the dependabot cooldown configuration and associated README documentation added in #560 (c778cb0).

## Changes

- **`.github/dependabot.yml`**: Remove the `cooldown.default` setting
- **`README.md`**: Remove the "Dependabot Cooldown Period" section and its table-of-contents entry

## Motivation

The cooldown feature introduced in #560 used the `default` key, which is not a valid option — Dependabot expects `default-days` instead. Rather than fixing the configuration in place, this PR reverts the cooldown changes entirely so we can re-evaluate whether a cooldown policy is the right approach for this repository.

> **Note:** The `.gitignore` expansion from #560 is intentionally kept; only the cooldown-related changes are reverted.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,5 +35,3 @@ updates:
       - dependency-name: "cpp-linter/cpp-linter-action"
         versions: ">=2.16"
     open-pull-requests-limit: 50
-    cooldown:
-      default-days: 7
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ This repository hosts GitHub Actions developed by the ASF community and approved
   - [Adding a New Action](#adding-a-new-action-to-the-allow-list)
   - [Reviewing](#reviewing)
   - [Adding a New Version](#adding-a-new-version-to-the-allow-list)
-    - [Dependabot Cooldown Period](#dependabot-cooldown-period)
   - [Manual Version Addition](#manual-addition-of-specific-versions)
   - [Removing a Version](#removing-a-version-manually)
 
@@ -143,23 +142,6 @@ This will:
 > [!NOTE]
 > **Prerequisites:** `docker`, `uv`, and `gh` (GitHub CLI, authenticated via `gh auth login`).
 > The build runs in a `node:20-slim` container so no local Node.js installation is needed.
-
-#### Dependabot Cooldown Period
-
-This repository uses a [Dependabot cooldown period](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#cooldown) of 7 days. After a Dependabot PR is merged or closed, Dependabot will wait 7 days before opening the next PR for the same ecosystem. This helps keep the volume of update PRs manageable and gives reviewers time to catch up.
-
-> [!TIP]
-> We recommend that ASF projects configure a similar cooldown in their own `dependabot.yml` to avoid being overwhelmed by update PRs and to catch up with approved actions here:
-> ```yaml
-> updates:
->   - package-ecosystem: "github-actions"
->     directory: "/"
->     schedule:
->       interval: "weekly"
->     cooldown:
->       default-days: 7
-> ```
-> Adjust the `default-days` value to match your project's review capacity.
 
 ### Manual Addition of Specific Versions
 


raboof (commented):
I would actually prefer not to have a cooldown period in *this* repository, because this dependabot workflow is also used for triggering reviews of updates to allowlisted GitHub Actions in `dummy.yml`. The quicker we can start reviewing the better, and these workflows are not actually ran, so there's no risk. Of course actual _projects_ may still apply a cooldown period. Having no cooldown period here will also reduce the risk that a GHA gets merged into a project *before* it gets reviewed and allowlisted here.

(sorry I missed #560)

potiuk:
I am not sure if we want to make it to 0. For sure we want to make it shorter than what is expected by projects (as long as we are sure there is no harm with having dependabot installing even malicious dependencies - theorethically withoug coldown - we might get malicious dependencies installed in PR.

It might cause some harm - potentially . Dependabot creates PR within this repo so token **might** be not read only (we have workflows with read), so potentialtly some secrets might leak for - might be not an issue for this repo though - but possibly I'd leave it - just to avoid reviewers having to review malicious code in case it will be updated by dependabot - and giving all kinds of analysis time to pull it even before we see it.

Also there is a potential that the reviewer will want to merge it quickly - and will miss malicious code and merges it so some cooldown is - I think -  needed.

potiuk:
> You're shifting goalposts here: first it's about security, with a highly questionable self-merge at #560 , now suddenly it's about saving reviewers' time.

Correct. I am. I just realized the second reason afterwards. Sometimes I do things intuitively, and realise consequences after discussion. This is what happened here. Still does not make it an invalid argument on it's own.

And this is in direct response to your question:

> I don't see reviewers reviewing malicious code as a problem - it's kinda the point?

Actually this question of yours, made me think that reviewers reviewing malicious code might simply add unnecesary friction/time/energy/frustration needed. 

And just a digression - We are all suffering now from too much of an overload because we receive PRs, issues, emails etc. that we need to take actions on. 

Anything that automatically decreases the likelihood of this is good. This how I optimize things in Airflow for years. Living the world a bit better every single day, and in this case my goal at least is to make those reviews easier, faster and with less unnecessary overhead.  And compound effect of it is actually enormous - but if I (And many other people now whom I lead by example there) - would not do it  - we would be light-years behind now and the amount of things to fix in devex would create unsurmountable problems.

I saw it at Snowflake, where DevEx was always an afterthought - and after almost 8 years of neglecting - when my firm was acquired, just the speed of using the development tools (not even the company processes) slowed us down 10x at least. Snowflake had invested a lot in it afterwards.

> Is this really a problem? I have not seen a single reviewer complain about having had to review a malicious update.

Not sure, but whenever I can I try to anticipate problems and fix them before they happen. I know it might be YAGNI -but still - that is very small change and has pretty much no negative effect, really - especially that we want everyone to use cooldowns (or so I think). This is my appraoch for all kind of DevEx - anticipate issues and fix them before they happen (of course also after they happen). I do it for years with quite a success.

See the compound problem above. Every time when I have a chance to improve DevEx - I do it - and when someone asks me if it already happened, I say, that - similarly as in case of getting ill, it's best not to get ill but prevent it. Especially if preventing costs practically 0.

> You're misrepresenting my position. I never said "review as soon as possible" is the goal .

I am not attacking you or misrepresenting your words, sorry if you took it this way. And I never said "you" said it. 
This is simply how I think and make logical conclusions.

If you ask me what kind of goals you might have when designing such review process, you might have several goals - and I simply think that naturally adding cooldown delays starting of release process. That's my thinking - not misrepresenting your words. So if I think: 0 or non-0, I see that 0 - allows to achieve the "review as fast as possible goal" while "1" achives "sometimes avoid review at all" goal (at the expense of when the review process started). And All I say, is for me the "sometimes avoid review at all" - is the goal I want by setting this value - because it helps to achieve different goal "make the review process as efficient and having as little overhead as possible" - that's definitely my primary goal.

My secondary goal - is to make it possible to distribute the review process so that many poople can do it - in case we have even more of those coming.

I am not sure what goals you have in mind for this process - but I thought that is one achieved by setting 0. 

> IMHO ideally it's less often then daily (per reviewer), but OK.

The old saying is "if things are painful - do them more often" and it works in many infrastructural kinds of work and all kinds of chores. Usually many small tasks done regularly spread over time, are much better than doing the same with seemingly unsurmountable number of those. But yes. It's OK to have them less frequently - it relly depends how much of it will happen in the future, how many PMCS we will have, how often they will be upgrading. I think it would be worth doing some calculations - based on practical observations in the future - especially if we will manage to either enforce or encourage people to use dependabot for the actions - that we will have more of those happening. Also that will make our reviews generally easier, because if you have to review next version change for an action rather than 10 next versions in one go, each of those changes is smaller and more localized, -  reviewing 10 lines PR is disproportionaly faster than 100 lines PR. From experience it seems that the time needed to review PR is roughly quadratic to the number of lines changes (or in some cases almost logarithmically if you have complex algorithms).

> I still maintain that delaying opening the PR for review in this repo is more harmful than helpful. It adds friction: 

What friction? It happens automatically - only a bit later after new version of an action has not been pulled (which of course happens rarely) - but I do not see any friction to be honest. It's unlikely someone using the action will update to new version quickly - and if they do, that's even a great opportunity to teach them that they also shoudl use - and even longer - cooldown. So if - in this case friction is added for the PMC, that's a good kind of friction - one we want, a friction that does not cause immediate problem with their current version - but make them pause and reach out - and learn that they should use dependabot. Goal of teaching dependabot achieved. 

I can't see any other friction here - only good one (And yes I strongly think that we should learn to use healthy friction to achieve our goals -  this seems to be contradicting with "increasing efficiency" - but actually a very deliberately and well though friction is a great way to achieve long-term efficience by introducing "pauses" where people have to think, discuss and as result -  improve their processes.


> regardless of the exact numbers, the longer you delay the review, the more people in projects will see dependabot PRs for actions that haven't been reviewed yet. Another way to say that is that it increases time pressure on the review, which I don't see a good reason for.

Not really - if they follow our cooldown of 4 days - that gives us 3 days to review it. If they have no cooldowns and are ahead of us, this is the learning opportunity - that they **should** use those 4 days of cooldown. And no - it does not increase the pressure. It shorten the pressure by 1 day. If we are going to suggest (or even enforce) cooldown on our PMCs, we can actually make that cooldown we suggest even longer. For example if our target review time is 4 days - we should suggest people to use 5 days cooldown. And if they come before - we should tell them, look, just set cooldown days now - do not upgrade the actions now and wait for review.

Really the difference is +1 day comparing to the cooldown we suggest to people - and we can suggest +1 more to achieve **exactly the same pressure** when it is set to 0.

Of course this does not work for people who do not have cooldown or have shorter cooldown - but then our response shoudl be "Set the cooldown to our suggested value" and that's - again - great way to teach our PMCs to set proper cooldown.

potiuk:
* Allowlist action added in `airflow` and merged: https://github.com/apache/airflow/pull/64450
* Improved your action's output in https://github.com/apache/infrastructure-actions/pull/630

I also created with Claude (This one took 20 minutes) that audits all the repos and creates PR adding missing security features:

* Dependabot with cooldown
* CodeQL scanning 
* Adding zizmor 
* Adding "allowlist-check" action

It will check what is existing, print a nice summary  of what is in the repo, what is missing, and whether it has already a PR opened by previous run of the script: https://github.com/apache/infrastructure-actions/pull/631

I still need to test it and will run in on a few repos if that sounds like a good idea.

PR #611, opened 2026-03-27
Fix: pin ruyaml with uv lockfile and 4-day cooldown

potiuk:
## Summary

- **Security fix**: Previously, `pip install ruyaml` in workflows with `contents:write` permission (`update_actions`, `update_dummy`, `remove_expired`) fetched the latest version from PyPI on every run. A compromised ruyaml release could exploit this to modify `actions.yml` and inject malicious allowed actions into the ASF organization-wide allowlist.
- **Fix**: Added `pyproject.toml` with `uv.lock` and a 4-day `exclude-newer` cooldown. All gateway workflows now use `pipx install uv` + `uv run` instead of `pip install ruyaml`, ensuring only locked, vetted versions are used.
- **Dependabot**: Added `uv` package ecosystem to `dependabot.yml` (with 4-day cooldown) to keep `uv.lock` updated via PRs.

## Test plan

- [ ] Verify `update_actions` workflow runs successfully on PR
- [ ] Verify `remove_expired` workflow runs successfully
- [ ] Verify `update_dummy` workflow runs successfully
- [ ] Verify `pytest` workflow passes
- [ ] Confirm dependabot picks up the new `uv` ecosystem

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,5 +7,10 @@ dependencies = [
     "ruyaml>=0.91.0",
 ]
 
+[dependency-groups]
+dev = [
+    "pytest",
+]
+
 [tool.uv]
 exclude-newer = "4 days"
--- a/uv.lock
+++ b/uv.lock
@@ -3,8 +3,17 @@ revision = 3
 requires-python = ">=3.13"
 
 [options]
-exclude-newer = "2026-03-23T23:26:24.032328Z"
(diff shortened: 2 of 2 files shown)


potiuk:
This is important to review and merge as this solves an existing security risk with the action itself @dfoulks1 @gmcdonald @raboof

PR #613, opened 2026-03-27
Add CI check for approved actions count approaching 1000 limit

### LINKED ISSUE #602: Approved Actions has a 1000 entry limit.
Currently, we're at 280/1000 actions in our org wide 'approved_patterns' list. This issue is to keep us aware of this limitation (it's not highly advertised) and to foster conversation about potential solutions and workarounds.

potiuk:
## Summary

- Adds a workflow that counts entries in `approved_patterns.yml` and fails at 800 (out of a hard limit of 1000 org-wide approved actions)
- Runs on pushes to main and PRs that modify `approved_patterns.yml`
- Gives clear error message explaining the situation and linking to #602

Currently at ~280 entries, so plenty of headroom, but this ensures we get early warning.

Closes #602

## Test plan

- [ ] Verify workflow runs on PR and passes (current count ~280, well under 800)
- [ ] Verify the error message is clear if threshold is exceeded

### DIFF
--- a/.github/workflows/check_approved_limit.yml
+++ b/.github/workflows/check_approved_limit.yml
@@ -1,0 +1,36 @@
+name: Check Approved Actions Limit
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - approved_patterns.yml
+  pull_request:
+    paths:
+      - approved_patterns.yml
+      - .github/workflows/check_approved_limit.yml
+
+permissions:
+  contents: read
+
+jobs:
+  check-limit:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/check_approved_limit.yml
```diff
@@ -0,0 +1,36 @@
+name: Check Approved Actions Limit
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - approved_patterns.yml
```
raboof: `approved_patterns.yml` should never be manually edited, so it makes no sense to have this trigger

Issue #600, opened 2026-03-26
verify-action-build: validate composite actions with minified javascript


raboof:
#592 updates a few actions under gradle/actions, and does have a `dist` folder that would be good to validate.

Currently it shows 

```
╭─────────────────────────── SKIPPED ───────────────────────────╮
│ This is a composite action, not a JavaScript action.          │
│ Build verification of compiled JS is not applicable.          │
╰───────────────────────────────────────────────────────────────╯
```

PR #617, opened 2026-03-28
Fix verify-action-build.py for monorepo and non-dist output dirs

potiuk:
Two issues fixed:

1. Actions using non-dist output directories (e.g. azure/setup-kubectl uses lib/) were not verified. The script now parses the main: field from action.yml to detect the correct output directory.

2. Monorepo actions like gradle/actions with sub-actions (setup-gradle, dependency-submission, wrapper-validation) failed because:
   - The root action.yml is composite, so verification was skipped
   - The build system lives in sources/ subdirectory, not at root
   - extract_action_ref_from_pr only returned the first ref

   Now the script:
   - Parses sub-paths from action refs (org/repo/sub@hash)
   - Extracts ALL action refs from PR diffs
   - Groups sub-actions by repo and verifies each one
   - Detects build directories (sources/, src/) when no root package.json
   - Tries root-level build scripts (e.g. ./build dist) first

Tested against:
- azure/setup-kubectl@776406b (lib/ output dir) - 1 JS file, identical
- gradle/actions/setup-gradle@39e147c (monorepo) - 5 JS files, all identical

- [ ] Compiled JavaScript in `dist/` matches a clean rebuild (verify with `uv run utils/verify-action-build.py org/repo@hash`)

Closes: #600

### DIFF
--- a/utils/verify-action-build.py
+++ b/utils/verify-action-build.py
@@ -39,8 +39,13 @@ output = Console()
 ACTIONS_YML = Path(__file__).resolve().parent.parent / "actions.yml"
 
 
-def parse_action_ref(ref: str) -> tuple[str, str, str]:
-    """Parse org/repo@hash into (org, repo, hash)."""
+def parse_action_ref(ref: str) -> tuple[str, str, str, str]:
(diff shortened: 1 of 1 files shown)


raboof:
This should fix #600, right?

PR #619, opened 2026-03-28
Fix `pypa/gh-action-pypi-publish`: pin to commit SHA instead of branch pattern

kevinjqliu:
`pypa/gh-action-pypi-publish` was pinned to `release/v1*` (a branch pattern) instead of an exact commit SHA, violating the [pinning policy](https://github.com/apache/infrastructure-actions#adding-a-new-action-to-the-allow-list).

Updated to pin to [`ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e`](https://github.com/pypa/gh-action-pypi-publish/commit/ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e) (v1.13.0).

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -675,7 +675,7 @@ pypa/gh-action-pip-audit:
     tag: v1.1.0
     expires_at: 2026-03-31
 pypa/gh-action-pypi-publish:
-  'ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e':
+  ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e:
     keep: true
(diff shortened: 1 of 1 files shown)


actions.yml
```diff
@@ -675,8 +675,9 @@ pypa/gh-action-pip-audit:
     tag: v1.1.0
     expires_at: 2026-03-31
 pypa/gh-action-pypi-publish:
-  release/v1*:
+  'ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e':
```
raboof: we don't typically use quotes here

PR #648, opened 2026-04-01
Prevent workflow command injection in verify action build output

potiuk:
## Summary

- Wrap `verify-action-build.py` invocation with `::stop-commands::<token>` / `::<token>::` using a random UUID
- Prevents strings like `##[add-matcher]`, `::error::`, `::warning::` in action diff output from being interpreted as GitHub Actions workflow commands

## Test plan

- [ ] Trigger the verify workflow on a dependabot PR and confirm diff output containing workflow-command-like strings is printed as plain text

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/verify_dependabot_action.yml
+++ b/.github/workflows/verify_dependabot_action.yml
@@ -39,6 +39,15 @@ jobs:
       - run: pipx install uv
 
       - name: Verify action build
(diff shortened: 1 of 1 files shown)


potiuk:
> I was wondering about those errors ...

I just asked AI - Copilot to diagnose and Claude to fix it.

PR #674, opened 2026-04-06
Retry JS verification with approved version's lock files on mismatch

potiuk:
When the initial JS build verification fails and a previously approved version exists, retry the Docker build using the approved commit's package.json and lock files.  This handles the common case where a version bump updates devDependencies (e.g. rollup, ncc) but the committed dist/ was built with the old toolchain.

If the retry matches, the result is reported as a warning ("matches with approved lock files, devDeps changed") rather than a hard failure, with guidance to review the devDependency changes before approving.

### DIFF
--- a/utils/verify_action_build/verification.py
+++ b/utils/verify_action_build/verification.py
@@ -119,9 +119,14 @@ def verify_single_action(
     """Verify a single action reference. Returns True if verification passed."""
     org, repo, sub_path, commit_hash = parse_action_ref(action_ref)
 
+    # Look up approved versions early — used for the lock-file retry and the
+    # later approved-version diff section.
+    approved = find_approved_versions(org, repo)
+
     checks_performed: list[tuple[str, str, str]] = []
     non_js_warnings: list[str] = []
     checked_actions: list[dict] = []
+    matched_with_approved_lockfile = False
 
     with tempfile.TemporaryDirectory(prefix="verify-action-") as tmp:
         work_dir = Path(tmp)
@@ -234,12 +239,9 @@ def verify_single_action(
             all_match = diff_js_files(
                 original_dir, rebuilt_dir, org, repo, commit_hash, out_dir_name,
             )
-            checks_performed.append((
-                "JS build verification",
-                "pass" if all_match else "fail",
-                "compiled JS matches rebuild" if all_match else "DIFFERENCES DETECTED",
-            ))
 
+            # If no compiled JS was found in dist/ but node_modules is vendored,
+            # verify node_modules instead
             if has_node_modules:
                 nm_match = diff_node_modules(
                     original_node_modules, rebuilt_node_modules,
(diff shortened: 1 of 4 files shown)


dave2wave (approved):
Looks reasonable to me, but wait for another review

potiuk:
Rebased after merging the split refactor. CC: @raboof -> that is a **weird** case but it happened already once that action's javascript was actually built before the dependencies from previous run where upgraded - resulting in non-consistent javascript (but it actually shows that our javascript reproducibility checks are **good** because essentially anyone could commit  **any** modification to compiled javascript - this one is at least reproducible with **previously approved** version of dependencies - so it is **somewhat reproducible**.

The funny thing - I would have probably never guessed that this happened, it was entirely figured out by Claude - I just asked it - what's wrong with this PR and can you produce a fix - and it did - with very reasonable explanation :)

potiuk:
Also I asked Claude - and it diagnosed the Sonar Qube action code and produced a fix that I submitted as PR https://github.com/SonarSource/sonarqube-scan-action/pull/228

dave2wave:
@potiuk So if someone is building their action in this way is this not a **_bug_** on their part and we should reject this action completely? Seems like they are producing a product that were it an ASF release it would be rejected.

potiuk:
> @potiuk So if someone is building their action in this way is this not a **_bug_** on their part and we should reject this action completely? Seems like they are producing a product that were it an ASF release it would be rejected.

We might - yes - depending on our criteria and goals

* if we care about "correctness" - yes, we should reject it
* if we care anout makicng it ready for people who want to do it fast and to only check safety - no, we should not 

We know it's a bit sloppy but should be safe (because it is reproducible). Especially if it is followed (which is what I did) with an issue or PR fixing it - this is I think the best approach - when we notice an issue and let the third party know so they fix it, and then continue with a workaround (which this check really is). 

They might not be super sloppy, they might made an innocent mistake in their pipeline (I think this is what happened in this case). I think we should be quite careful what we "punish" by not accepting actions - because we essentially punish our PMCs.  And I find it much more open-source friendly to work with our dependencies rather than reject them because they made a mistake - and even do not tell them that they did. If we don't tell them, they might not ever be aware of it.

This is also folllowing what Astral team explained in the https://astral.sh/blog/open-source-security-at-astral astral, we could have similar approach as them). They worked with they downstream dependencies to make sure all their composite dependencies are hashed - which likely should also be our criteria - I was thinking about adding such criteria. This **might** be a good reason for rejection (because of security)

My personal goal with current way we handle actions (and this is why I developed verify_build_actions.p) was to unblock our PMCs as soon as possible, while making sure we are safe. I did not even have the "licence" criteria in mind (this is why I missed it when reviewing Gradle, but that is another one to consider, if it should be our review criteria as well).

If we would like to add "correctness" of building the action and reject those actions based on it, we can - but It was not my primary concern so far, and for me "freeing" actions so that our PMCs can upgrade to latest "safe" ones as fast as possible (minus cooldown for security reasons) was the primary criteria (and the only so far).

What's your criteria for accepting / rejecting the actions @dave2wave - maybe we have different ones?

potiuk:
I created https://github.com/apache/infrastructure-actions/issues/686 -> maybe a discussion can be held there.

raboof:
> if someone is building their action in this way is this not a bug on their part

I'd say so

> and we should reject this action completely?

I'm OK with approving #669 in this case: our goal is to make sure all actions are non-malicious, not that they're perfect.

That said, I'm not sure we should add this feature to the check script: I think the failure is a useful reminder that 'something is wrong upstream'. I love that @potiuk proposed a fix upstream so ideally it wouldn't happen again next time. If the problem remains that might become a reason to migrate away from this action eventually. The more of this complexity we add to the check tool, the harder it is to understand and thus trust its output.

So I think we should probably drop this PR

potiuk:
> So I think we should probably drop this PR

I think this case is likely to happen in other cases - because it's very easy to make similar mistake like they did

For now it actually produces warning if this case happen, so maybe it's better to leave it - otherwise when it happens next time we will not know when it happens.

potiuk:
> For now it actually produces warning if this case happen, so maybe it's better to leave it - otherwise when it happens next time we will not know when it happens.

Optionally, we can get it **fail** rather than raise warning - but we should at least know that we had similar case.

utils/verify_action_build/verification.py
```diff
@@ -234,21 +238,86 @@ def verify_single_action(
...
+                            "before approving.",
+                            border_style="yellow",
+                            title="MATCHED WITH APPROVED LOCK FILES",
+                        )
+                    )
+
+            checks_performed.append((
+                "JS build verification",
+                "pass" if all_match and not used_approved_lockfile else
+                "warn" if all_match and used_approved_lockfile else "fail",
```
dave2wave: Is this where to apply the choice to fail or warn?

potiuk: I just updated it - the updated change has:

* regresssion tests
* fails instead of warning when we detect previously approved lock files (with detailed information)l

potiuk: Here is the fail: https://github.com/apache/infrastructure-actions/pull/674/changes#diff-7a5c5e11a83966f00b690dfa76753e32949bc18bb7709fda53713449f8410251R315

potiuk:
@raboof - do you think - failing check with clear info is good enough ?

potiuk:
FYI:  https://community.sonarsource.com/t/the-sonarsource-sonarqube-scan-action-7-1-0-build-output-is-not-reproducible/180749/2  there is a discord session on sonarsource - apparently someone else **also** does reproducibility checks on their actions.


Seeing that - It's **really** good idea I do it, this is an attack vector that nobody talks about but I am pretty sure someone will exploit it soon. It's extremely easy to pull off. The Attestations by @ppkarwasz might also be a good one to add (let's see what investigation on how many of our actions could be verified this way: https://github.com/apache/infrastructure-actions/issues/686#issuecomment-4271063001 - having attestations might be useful signal to add. 

@raboof -> I am merging this one for now - it will try to do such "last good deps" check and fail if it finds this is the case but at least we will know the exact reason for lack of reproducibility.

PR #685, opened 2026-04-09
Fix Docker build for actions with multi-step builds

potiuk:
## Summary
- Fixes the verify-action-build Docker rebuild for actions that need multiple build steps (e.g. `tsc` → `lib/` then `ncc build` → `dist/`)
- The previous `elif` chain stopped after the first successful build command even if it didn't produce output in the expected directory
- Replaced with a loop that checks for JS output after each step and continues until `dist/` has files
- This unblocks #667 and #668 (matlab-actions dependabot bumps) which fail verification because `npm run build` (tsc) succeeds but outputs to `lib/`, not `dist/`

## Verified
- Tested locally: `uv run utils/verify-action-build.py matlab-actions/setup-matlab@c6998d116f15` now passes (both `cache-save/index.js` and `setup/index.js` show as identical)
- All 112 existing unit tests pass

## Test plan
- [ ] CI passes
- [ ] Re-run verify on #667 and #668 after merging

### DIFF
--- a/utils/verify_action_build/dockerfiles/build_action.Dockerfile
+++ b/utils/verify_action_build/dockerfiles/build_action.Dockerfile
@@ -153,33 +153,45 @@ RUN BUILD_DIR=$(cat /build-dir.txt); \
     fi
 
 # Build: first try a root-level build script (some repos like gradle/actions use one),
-# then try npm/yarn/pnpm build in the build directory, then package, then start, then ncc fallback.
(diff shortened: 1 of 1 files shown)


potiuk:
Indeed - that one nicely worked for #667 and #668

raboof (approved):
OK, though this also is starting to reach the complexity where regression tests start to make sense

PR #696, opened 2026-04-12
Re-add missed allowlist additions

raboof:
These were missed in the period that branch protection prevented the workflow from copying allowlists from the dummy workflow to `actions.yml`.

(#689 added them to `approved_patterns.yml`, but once the automated jobs were running again they were removed again because they didn't show up in `actions.yml`)

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -280,6 +280,9 @@ docker/login-action:
     expires_at: 2026-06-14
   b45d80f862d83dbcd57f89517bcf500b2ab88fb2:
     tag: v4.0.0
+    expires_at: 2026-09-14
+  4907a6ddec9925e35a0a9e82d7399ccc52663121:
+    tag: v4.1.0
 docker/metadata-action:
   c299e40c65443455700f0fdfc63efafe5b349051:
     tag: v5.10.0
@@ -349,6 +352,9 @@ erisu/apache-rat-action:
 erisu/license-checker-action:
   99cffa11264fe545fd0baa6c13bca5a00ae608f2:
     tag: v2.0.1
+    expires_at: 2026-09-14
+  04511f4c052b5773f11e1c65b42cda88235c62ae:
+    tag: v9.2.0
 geekyeggo/delete-artifact:
   '*':
     keep: true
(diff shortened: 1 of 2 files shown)


actions.yml
```diff
@@ -349,6 +352,9 @@ erisu/apache-rat-action:
 erisu/license-checker-action:
   99cffa11264fe545fd0baa6c13bca5a00ae608f2:
     tag: v2.0.1
+    expires_at: 2026-09-14
+  04511f4c052b5773f11e1c65b42cda88235c62ae:
+    tag: v9.2.0
```
erisu: The current tag should be `v2.1.0`. 

Release Notes: https://github.com/erisu/license-checker-action/releases/tag/v2.1.0

The dependabot PR that updated to v2.1.0: https://github.com/apache/infrastructure-actions/pull/680

```suggestion
    tag: v2.1.0
```

erisu (changes requested):
It appears that the new tag version for `erisu/license-checker-action` has been written incorrectly.

potiuk (approved):

actions.yml
```diff
@@ -537,6 +552,9 @@ JustinBeckwith/linkinator-action:
     expires_at: 2026-05-22
   f62ba0c110a76effb2ee6022cc6ce4ab161085e3:
     tag: v2.4
+    expires_at: 2026-07-05
+  363572b2714d25a059fceb2fa332a98e7ea3baff:
+    tag: v2.4
```
dave2wave, line 557: This should also be 2.4.1

raboof: good observation, #698

potiuk: Ah - I missed that @raboof  modified the dummy-workflow.yml.

So I copied the output from the last failed "Handle Dependabot Upgrade"  https://github.com/apache/infrastructure-actions/actions/runs/24309558948/job/70976723553 

The "actions.yml" and "approved-actions.yml" files are generated automatically from dependabot.yml -> this is why I have not generated it in the first place (so @raboof @dave2wave.


So really this one here was not "missing entries" - the updates  were just not pushed by the "Handle Dependabot Update" workflow - because we did not have branch protection removed yet) - until

In https://github.com/apache/infrastructure-actions/pull/689 I only added "approved actions" to unblock people but I wanted to let the "Handle Dependabot Upgrade" do the job and update "actions.yml" automatically - yes, it was lazy likely.

raboof: > In https://github.com/apache/infrastructure-actions/pull/689 I only added "approved actions" to unblock people but I wanted to let the "Handle Dependabot Upgrade" do the job and update "actions.yml" automatically - yes, it was lazy likely.

Not sure if 'lazy' is the word, you were quite correct that ideally `actions.yml` would have been automatically updated based on `dummy.yml` - but since that wasn't triggered back when branch protection was still enabled I figured I'd 'help it along' with this PR.

PR #692, opened 2026-04-10
Revert branch protection #642

dave2wave:
This comments out the branch protection of `main`. Note that INFRA may need to act if the merging of this does not remove branch protection.

PR #697, opened 2026-04-12
Leaving empty protected branches to actually disable it

potiuk:
The protected branches have not been removed by #692 because it just skipped the whole protected brnaches. Empty protected branches should cleanup all of them.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -20,8 +20,8 @@ github:
   del_branch_on_merge: true
   features:
     issues: true
+  protected_branches:
   # revert #642
-  #protected_branches:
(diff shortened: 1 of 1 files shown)


dave2wave (approved):
LGTM

raboof:
> The protected branches have not been removed by #692 because it just skipped the whole protected brnaches. Empty protected branches should cleanup all of them.

If that were the case then https://github.com/apache/infrastructure-actions/commit/0233fe3d03e3b4104ac09b284c2a267130f9580a would not have been possible, right?

No objection to the change, though.

Issue #606, opened 2026-03-26
Better diagnostics for un-allowlisted transitive actions


raboof:
Since #279, the `dummy.yml` workflow is regularly invoked.

The goal of that was to detect the scenario where one of the listed actions is a composite action that depends on another action that has not been allowlisted yet (or is not allowlisted anymore). This is useful, but has a number of shortcomings:

* In case of new actions, this validation only occurs *after* the addition of an action with this problem has been merged. It would be great if we could perform this check as part of PR validation of new actions (and action updates). This is not a trivial matter of enabling the workflow on PRs, because it checks the listed actions against the allowlist actually configured in GitHub, which would not have the new action yet.
* If this workflow identifies a transitive non-allowlisted action, it's not always obvious to tell which action it's a transitive dependency of.
* Only the 'latest' version of each action is included in `dummy.yml`, so earlier still-allowlisted actions are not validated against their dependencies expiring

It would be really cool if we had our own script to check this, which could work for PRs and provide better diagnostics.

PR #712, opened 2026-04-14
dependabot: apply cooldown except for action reviews

### LINKED ISSUE #683: Determine recommendations for cooldown periods

raboof:
Fixes #683

Sadly GHA are not so easy to test, so this might take some rounds to get right :)


potiuk:
> Sadly GHA are not so easy to test, so this might take some rounds to get right :)

Try claude.... (and not joking about it) - for things like renaming workflows etc. It's actually very good in findng all the places that needs to get updated :)

raboof:
Fixed a few things, now seems OK to me.

That "Verify Dependabot Action Build" is skipped makes sense since this is not a dependabot PR
That "Check for transitive failures in current latest actions" fails is expected, that's #606

gateway/gateway.py
```diff
@@ -166,31 +166,21 @@ def generate_workflow(actions: ActionsYAML) -> str:
...
-  dummy:
-    runs-on: ubuntu-latest
-    steps:
+# Sadly the error message does not tell you *which* action
+# has a missing transitive dependency, see
+# https://github.com/apache/infrastructure-actions/issues/606
+name: Gateway Action
+
+runs:
+  steps:
```
potiuk, line 184: **Blocker:** the header literal is missing `using: "composite"`, so the first time `update_composite_actions.yml` runs it will regenerate the file without that line and GitHub will refuse to load it as a composite action. Worth pairing this with a tiny snapshot test (e.g. `assert 'using: "composite"' in generate_composite_action({})`) so this can't regress silently.

```suggestion
runs:
  using: "composite"
  steps:
```

gateway/gateway.py
```diff
@@ -256,16 +246,16 @@ def update_refs(
...
 
-def update_actions(dummy_path: Path, actions_path: Path):
+def update_actions(composite_action_path: Path, actions_path: Path):
     """
-    Update actions file based on a dummy workflow.
+    Update actions file based on the composite actions.
 
     Args:
-        dummy_path: Path to the dummy workflow file
+        composite_action_path: Path to the dummy workflow file
```
potiuk: Stale docstring — this parameter no longer points at a dummy workflow file.

```suggestion
        composite_action_path: Path to the composite action file
```

.github/workflows/update_composite_actions.yml
```diff
@@ -26,7 +26,7 @@ on:
       - "actions.yml"
   pull_request:
     paths:
-      - ".github/workflows/update_dummy.yml"
+      - ".github/workflows/update_composite_action.yml"
```
potiuk, line 29: **Blocker / typo:** the file is `update_composite_actions.yml` (plural), but the `paths:` filter references `update_composite_action.yml` (singular), so editing this workflow itself will no longer re-trigger it on PRs.

```suggestion
      - ".github/workflows/update_composite_actions.yml"
```

.github/workflows/update_composite_actions.yml
```diff
@@ -85,6 +85,6 @@ jobs:
           AUTHOR_EMAIL=$(gh api /user --jq '.email // "\(.login)@users.noreply.github.com"' 2>/dev/null || echo "asfgit@users.noreply.github.com")
           git config --local user.name "${AUTHOR_NAME}"
           git config --local user.email "${AUTHOR_EMAIL}"
-          git add -f .github/workflows/dummy.yml approved_patterns.yml
-          git commit -m "Update approved_patterns.yml and .github/workflows/dummy.yml based on actions.yml" -m "Generated by .github/workflows/update_dummy.yml" || echo "No changes"
+          git add -f .github/actions/for-dependabot-triggered-reviews/action.yml approved_patterns.yml
+          git commit -m "Update approved_patterns.yml and .github/actions/for-dependabot-triggered-reviews/action.yml based on actions.yml" -m "Generated by .github/workflows/update_composite_action.yml" || echo "No changes"
```
potiuk: Same singular/plural mismatch in the `-m` metadata on the commit.

```suggestion
          git commit -m "Update approved_patterns.yml and .github/actions/for-dependabot-triggered-reviews/action.yml based on actions.yml" -m "Generated by .github/workflows/update_composite_actions.yml" || echo "No changes"
```

.github/dependabot.yml
```diff
@@ -34,14 +31,24 @@ updates:
       # https://github.com/apache/infrastructure-actions/issues/324
       - dependency-name: "cpp-linter/cpp-linter-action"
         versions: ">=2.16"
```
potiuk, line 33: **Blocker:** `open-pull-requests-limit: 50` currently sits on the *second* update block, but the 90+-entry allowlist-review surface is on *this* block and will otherwise fall back to dependabot's default of 5 PRs, silently throttling reviews. Hoist it up to this block…

```suggestion
        versions: ">=2.16"
    open-pull-requests-limit: 50
```

.github/dependabot.yml
```diff
@@ -34,14 +31,24 @@ updates:
...
+    schedule:
+      # 'daily' only runs on weekdays
+      interval: "cron"
+      cronjob: "45 13 * * *"
+    directories:
+      - "/.github/workflows"
+      - "/pelican/"
+      - "/stash/restore"
+      - "/stash/save"
     open-pull-requests-limit: 50
```
potiuk: …and drop it from this block, which only covers the real workflow / pelican / stash paths and doesn't need the high limit.

```suggestion
      - "/stash/save"
```

potiuk (commented):
Following up on my earlier review — turning the actionable items into inline suggestions. Three of these (the `using: "composite"` header, the `update_composite_action.yml` typo, and the `open-pull-requests-limit: 50` placement) are blockers in my opinion; the rest are polish.

potiuk (commented):
Re-review on `e488982` — all three blockers and every polish item from the previous pass are addressed. Confirmed against the new HEAD:

**Blockers**
- ✅ `gateway/gateway.py:183` — `using: "composite"` is now in the header literal.
- ✅ `update_composite_action.yml` — the singular/plural mismatch is resolved by renaming the file (the opposite direction from my suggestion, which works just as well since the in-file references are now consistent).
- ✅ `.github/dependabot.yml` — `open-pull-requests-limit: 50` moved onto the allowlist-review block.

**Polish**
- ✅ `update_refs`: `dummy_steps` → `composite_steps` (signature + docstring + body).
- ✅ `update_actions`: docstring `Path to the dummy workflow file` → `Path to the composite action file`.
- ✅ `update_workflow`: `dummy_path` → `composite_action_path`, docstring updated, and `write_str` call fixed to match.
- ✅ Bonus: `name: Update Dummy workflow` → `name: Update Approved Patterns and Composite Action`.

**Still open (non-blocking, from the prose review)**
- `cpp-linter/cpp-linter-action >=2.16` `ignore:` is only on the allowlist-review block. In practice it's fine today (the action isn't `uses:`-d anywhere in the real workflow tree), but if that ever changes the ignore won't apply — worth a durable comment or a duplicated `ignore:` on the second block.
- The `leafo/gh-actions-luarocks` re-ordering in `actions.yml` is unrelated to the cooldown refactor. Consider splitting it out for a cleaner history, or at least calling it out in the PR description.
- Optional: a one-line snapshot test asserting `'using: "composite"' in generate_composite_action({})` so the regression caught here cannot sneak back in.

LGTM apart from those nits.

potiuk (approved):

PR #715, opened 2026-04-14
stash/restore: add retry-count and fail-on-download inputs

potiuk:
## Summary

- Add `retry-count` input (default `3`) to `stash/restore/action.yml`. When `gh run download` exits with code `1` (the transient failure mode observed for artifact downloads), the download is retried up to `retry-count` times. Other exit codes are not retried.
- Add `fail-on-download` input (default `false`). When `true`, the step fails if the stash artifact was found but could not be downloaded after exhausting retries. When `false`, the failure is reported only via the `stash-hit` output.
- Harden the Download Stash script against `set -eo pipefail` (the default GitHub Actions bash shell): call `set +e` up front so a single failing command cannot abort the step, and handle exit codes explicitly via `$?`.
- Fix the previous `|| download="failed" && download="success"` one-liner, which due to `||`/`&&` precedence always reported `success` regardless of whether the download actually worked.

## Test plan

- [ ] Trigger a workflow that uses `stash/restore` against a valid stash and confirm the artifact is downloaded on the first attempt and `stash-hit` is `true`.
- [ ] Simulate a transient `gh run download` failure (exit 1) and confirm the action retries up to `retry-count` times.
- [ ] Run with `fail-on-download: true` against a known-broken download and confirm the step fails with the error annotation.
- [ ] Run with `fail-on-download: false` (default) against a known-broken download and confirm the step succeeds with `stash-hit=false`.

### DIFF
--- a/stash/restore/action.yml
+++ b/stash/restore/action.yml
@@ -45,6 +45,20 @@ inputs:
         If true, only the current branch will be searched for the stash.
         If false, the base branch(PRs)/default branch will be searched as well.
     default: "false"
+  retry-count:
+    description: >
+      Number of attempts for downloading the stash artifact. When `gh run download`
+      exits with code 1 (the transient failure mode observed for artifact downloads),
+      the download will be retried until it succeeds or this many attempts have been
+      made. Other exit codes are not retried.
+    default: "3"
+  fail-on-download:
+    description: >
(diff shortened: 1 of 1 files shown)


potiuk:
will open it from main repo to test

PR #716, opened 2026-04-14
stash/restore: add retry-count and fail-on-download inputs

potiuk:
## Summary

- Add `retry-count` input (default `3`) to `stash/restore/action.yml`. When `gh run download` exits with code `1` (the transient failure mode observed for artifact downloads), the download is retried up to `retry-count` times. Other exit codes are not retried.
- Add `fail-on-download` input (default `false`). When `true`, the step fails if the stash artifact was found but could not be downloaded after exhausting retries. When `false`, the failure is reported only via the `stash-hit` output.
- Fix the previous `|| download="failed" && download="success"` one-liner in the Download Stash step, which due to `||`/`&&` precedence always reported `success` regardless of whether the download actually worked.
- Move the retry / clean / fail-on-download logic out of inline bash into a `stash/restore/download_stash.py` module that exposes a single `download_stash(env, run_download=...)` function. The Download Stash step now runs it via `shell: python3 {0}`, matching the style of the existing Check for stash artifact step.
- Add `stash/restore/test_download_stash.py` (`unittest`) with a `FakeGh` stub injected through the `run_download` parameter so the logic can be exercised offline. Cases covered:
  - success on first attempt
  - retry on exit 1 until success
  - all retries exhausted with `fail-on-download=false` (tolerated, exit 0)
  - all retries exhausted with `fail-on-download=true` (exit 1, error annotation)
  - non-1 exit codes are not retried
  - `CLEAN=true` removes the stash dir before download
  - `CLEAN=false` leaves the stash dir contents alone
  - mixed transient+fatal stops on the first non-1 code
- Wire the new Python tests into the `stash-action-test` workflow next to the existing `test_get_stash.py` run, so they execute on ubuntu / windows / macOS.

## Test plan

- [x] `python3 stash/restore/test_download_stash.py` passes locally (8/8 cases).
- [x] `stash-action-test` CI workflow runs the new Python tests on ubuntu / windows / macOS.
- [x] Trigger a workflow that uses `stash/restore` against a valid stash and confirm the artifact is downloaded and `stash-hit` is `true`.
- [x] Run with `fail-on-download: true` against a broken download and confirm the step fails with the error annotation.
- [x] Run with `fail-on-download: false` (default) against a broken download and confirm the step succeeds with `stash-hit=false`.

Supersedes #715.


potiuk:
Tested in https://github.com/apache/airflow/pull/65268 , linting fixed separately in #717

bugraoz93 (approved):

potiuk:
Can I get approval on that one ? Ths one will make our life way easier in Airflow if we can switch to the new version of airflow @dave2wave @dfoulks1

dave2wave (approved):
LGTM

PR #717, opened 2026-04-14
pelican: install lint deps from pyproject.toml

potiuk:
## Summary

The \`Linting and MyPy (Pelican)\` workflow fails on every run because its Install dependencies step does:

\`\`\`
cd pelican
pip install -r requirements.txt --no-deps
\`\`\`

but \`pelican/requirements.txt\` no longer exists — pelican is declared via PEP 621 \`pelican/pyproject.toml\` these days. Replace the broken install with \`pip install ./pelican\`, which pulls in the runtime deps declared under \`[project.dependencies]\`. The mypy / pylint type-stub install line is unchanged; those are still needed on top of the runtime deps for the type-checking and pylint steps that follow.

## Test plan

- [ ] Push triggers the lint workflow and it reaches the \`Type testing with mypy\` and \`Code linting with pylint\` steps (no install-step failure).
- [ ] \`mypy --ignore-missing-imports .\` and \`pylint **/*.py\` still run against the pelican tree as before.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -45,24 +45,12 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57  # v8.0.0
 
-      - name: Install lint / mypy dependencies
-        run: |
-          # `--no-install-project` skips building `apache-pelican-action`
-          # itself; we only need the runtime dependencies plus the `dev`
-          # group (mypy, pylint and type stubs) to run the checks below.
-          cd pelican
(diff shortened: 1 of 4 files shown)


bugraoz93 (approved):

potiuk:
I'd love to merge it - just linting issue :)

potiuk:
(why we need it) That will actually help us in Airflow to get rid of some flaky tests - this one blocks the  https://github.com/apache/infrastructure-actions/pull/716 (and that one I would love to merge too).

dave2wave (approved):

PR #723, opened 2026-04-15
pin apache/infrastructure-actions/pelican to 93dfe4693bc118397840e7e4ae447e57a3eea7ee

kevinjqliu:
Requested by ASF Infra to pin to [93dfe4693bc118397840e7e4ae447e57a3eea7ee](https://github.com/apache/infrastructure-actions/commit/93dfe4693bc118397840e7e4ae447e57a3eea7ee)
Devlist thread https://lists.apache.org/thread/zd9cpkhwt0xh2sq958gwc62cw4f86yc7

We should pin to hash commit in the examples in this repo too

### DIFF
--- a/pelican/README.md
+++ b/pelican/README.md
@@ -45,7 +45,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: apache/infrastructure-actions/pelican@main
+      - uses: apache/infrastructure-actions/pelican@93dfe4693bc118397840e7e4ae447e57a3eea7ee # main
         with:
           destination: master
           gfm: 'true'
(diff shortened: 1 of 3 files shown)


dave2wave:
You misunderstood my email. The email was meant to imply you should pin it IF you had problems.

kevinjqliu:
oops ive been working with zizmor and pinning all the github workflows to commit hash, that's where my mind went automatically 🤣 

do you think its still valuable to pin to a commit hash here? otherwise, happy to close the PR

dave2wave (changes requested):
Please add commentary that builds can either be `main` or pinned to a hash like this one, but then if you pin you will need to update as we roll out updates and we update the actions used by our pelican action.

The email sent was a warning that you had problems you should report them and then pin.

dave2wave:
If you pin and don't have to then you will have to follow updates and it's possible your build will start to silently break because the pelican action you pinned to may be using an expired action. The pelican action post the hash uses uv and that setup action's hash will eventually expire.

kevinjqliu:
gotcha, i think we can just close this PR for now. thanks!

PR #725, opened 2026-04-16
Improve CI security posture: add CodeQL & zizmor workflows, fix dependabot config

kevinjqliu:
## Changes

- **Add CodeQL workflow** for GitHub Actions security scanning (`codeql.yml`)
- **Add zizmor workflow** for Actions-specific static analysis (`zizmor.yml`)
- **Fix dependabot config**: correct `versions` field to use list syntax, add license header formatting, add `zizmor: ignore` annotation for cooldown rule
- **Fix actor check** in `verify_dependabot_action.yml`: use `github.event.pull_request.user.login` instead of `github.actor` to avoid false negatives when non-dependabot users interact with the PR
- **Pin `matlab-actions/run-tests`** to v3.0.0 (zizmor fix)
- **Harden `pelican/action.yml`**: replace inline `${{ inputs.* }}` expressions in `run:` blocks with environment variables to prevent script injection
- **Harden `stash/restore/action.yml`**: move `${{ inputs.clean }}` into an `INPUTS_CLEAN` env var for the same reason
- **Pin `actions/upload-artifact`** in `stash/save/action.yml` to full SHA (`v7`)

Validated locally
```
GH_TOKEN=`gh auth token` uvx zizmor --min-severity medium --min-confidence medium .github/ --fix=all
```

### DIFF
--- a/.github/actions/for-dependabot-triggered-reviews/action.yml
+++ b/.github/actions/for-dependabot-triggered-reviews/action.yml
@@ -169,7 +169,7 @@ runs:
       if: false
     - uses: manusa/actions-setup-minikube@96202dee4ae1c2f46a62fe197273aaf22b83f42d  # v2.16.1
       if: false
-    - uses: matlab-actions/run-tests@353aee49b0edf62278c118a51b484d90bf6da1b7  # v3.1.0
+    - uses: matlab-actions/run-tests@4be3345d41da8b1bf8cc5cb01a5e19c4611cb15d # v3.0.0
       if: false
     - uses: matlab-actions/setup-matlab@a0180c939fb1a28de13f44f7b778b912384ced1f  # v3.0.1
       if: false
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,7 +19,7 @@
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
+  - package-ecosystem: "github-actions" # zizmor: ignore[dependabot-cooldown] see #683 and #712
     commit-message:
       prefix: "action-allowlist-review"
     directories:
@@ -35,10 +35,7 @@ updates:
       - dependency-name: "cpp-linter/cpp-linter-action"
         versions:
           - ">=2.16"
-        # zizmor: ignore[dependabot-cooldown] see #683 and #712
     open-pull-requests-limit: 50
-    cooldown:
-      default-days: 7
   - package-ecosystem: "github-actions"
     schedule:
       # 'daily' only runs on weekdays
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,4 @@
+#
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -14,6 +15,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+#
+
 version: 2
 updates:
   - package-ecosystem: "github-actions"
@@ -30,8 +33,11 @@ updates:
       # we're not comfortable trusting fully yet. See also
       # https://github.com/apache/infrastructure-actions/issues/324
       - dependency-name: "cpp-linter/cpp-linter-action"
-        versions: ">=2.16"
+        versions:
+          - ">=2.16"
     open-pull-requests-limit: 50
+    cooldown:
+      default-days: 7
   - package-ecosystem: "github-actions"
     schedule:
       # 'daily' only runs on weekdays
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,6 +35,7 @@ updates:
       - dependency-name: "cpp-linter/cpp-linter-action"
         versions:
           - ">=2.16"
+        # zizmor: ignore[dependabot-cooldown] see #683 and #712
     open-pull-requests-limit: 50
     cooldown:
       default-days: 7
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -16,38 +16,31 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-name: Verify Dependabot Action Build
+
+name: GitHub Actions Security Analysis with zizmor 🌈
 
 on:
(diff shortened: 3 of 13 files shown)


.github/actions/for-dependabot-triggered-reviews/action.yml
```diff
@@ -169,7 +169,7 @@ runs:
       if: false
     - uses: manusa/actions-setup-minikube@96202dee4ae1c2f46a62fe197273aaf22b83f42d  # v2.16.1
       if: false
-    - uses: matlab-actions/run-tests@353aee49b0edf62278c118a51b484d90bf6da1b7  # v3.1.0
```
kevinjqliu, line 172: `353aee49b0edf62278c118a51b484d90bf6da1b7` is orphaned 
https://github.com/matlab-actions/run-tests/commit/353aee49b0edf62278c118a51b484d90bf6da1b7

v3.1.0 points to 64426f612825c99055c2b4887c08c135ab01be0b
https://github.com/matlab-actions/run-tests/releases/tag/v3.1.0

changing this to use v3.0.0 https://github.com/matlab-actions/run-tests/releases/tag/v3.0.0

.github/dependabot.yml
```diff
@@ -30,8 +33,11 @@ updates:
       # we're not comfortable trusting fully yet. See also
       # https://github.com/apache/infrastructure-actions/issues/324
       - dependency-name: "cpp-linter/cpp-linter-action"
-        versions: ">=2.16"
+        versions:
+          - ">=2.16"
```
kevinjqliu, line 37: versions is a list
https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#ignoring-specific-versions-or-ranges-of-versions

zizmor caught this locally

.github/dependabot.yml
```diff
@@ -30,8 +33,11 @@ updates:
...
     open-pull-requests-limit: 50
+    cooldown:
```
dave2wave: No - please see #683 and #712

dave2wave (dismissed):

.github/dependabot.yml
kevinjqliu: now i remember that convo. this was automatically added by zizmor. let me back out of it

kevinjqliu: removed!

pelican/action.yml
```diff
@@ -71,19 +71,19 @@ runs:
       # Install needs to run in separate shell so stdout is restored
       run: |
         (
-          test "${{ inputs.debug }}" == 'true' || exec >/dev/null
+          test "${INPUTS_DEBUG}" == 'true' || exec >/dev/null
```
dave2wave, line 74: What's going on with turning action parameters into environment variables?

dave2wave:
> Harden pelican/action.yml: replace inline ${{ inputs.* }} expressions in run: blocks with environment variables to prevent script injection
> Harden stash/restore/action.yml: move ${{ inputs.clean }} into an INPUTS_CLEAN env var for the same reason

These changes are breaking unless the environment variables are initialized from the action's inputs.

pelican/action.yml
potiuk: This is security feature. "${{ inputs.debug }}" is interpolated before bash command gets executed. Someone who can pass input from outside could set inputs.debug = '   "; cat /etc/passwd; test " ' and it would cause command injection. 

This is basically the same class of issues as SQL injection. Passing all inputs and potentially user supplied values to env variables and using those env variables is safe because unlike GitHub actions expression language, env variables are interpolated by bash, and when it happens it's already toi late to inject `;` and similar separators.

potiuk: https://docs.zizmor.sh/audits/#template-injection has all details

potiuk:
> > Harden pelican/action.yml: replace inline ${{ inputs.* }} expressions in run: blocks with environment variables to prevent script injection
> > Harden stash/restore/action.yml: move ${{ inputs.clean }} into an INPUTS_CLEAN env var for the same reason
> 
> These changes are breaking unless the environment variables are initialized from the action's inputs.

No. All inputs are also automatically available as `INPUT_`  env variables https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepswith

.github/actions/for-dependabot-triggered-reviews/action.yml
potiuk: Whoa. This is worrying. Is it not a sign that smth is wrong  ?

.github/workflows/verify_dependabot_action.yml
```diff
@@ -29,7 +29,7 @@ permissions:
 
 jobs:
   verify:
-    if: github.actor == 'dependabot[bot]'
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
```
potiuk, line 32: Good catch!

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,46 @@
+#
```
potiuk, line 1: Oh ... we did not have zizmor ???? Ups

stash/save/action.yml
```diff
@@ -120,7 +120,7 @@ runs:
 
     - name: Upload Stash
       id: upload
-      uses: actions/upload-artifact@v7
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
```
potiuk, line 123: Yeah. I also prefer to have "all" actions hashed. It's good for consistency.

potiuk (approved):
Looks good. But I think we need to look at the matlab things - were it came from.

potiuk:
Ah - it seems we have pre-release version updatred by dependabot. We should disable pre-release upates.


image

potiuk:
I think they changed release to "pre-release".

potiuk:
@kevinjqliu -> the zizmor issues should be silenced. In this case we do not want to create separate env.

pelican/action.yml
dave2wave: Thanks!

kevinjqliu:
ci is green!

potiuk (approved):

potiuk:
Nice!

PR #733, opened 2026-04-18
verify-action-build: skip pin warning for multi-stage FROM references

potiuk:
## Summary
- `analyze_dockerfile` flagged `FROM <alias>` as unpinned when the alias was defined by an earlier `FROM ... AS <alias>` in the same Dockerfile. Multi-stage internal references don't pull an image, so the warning was noise.
- Track aliases introduced by `FROM <image> AS <alias>` and treat a later `FROM` whose image matches a known alias as a multi-stage reference instead of emitting a pin warning.
- Tightened the FROM regex from `(.+?)` to `(\S+)` so the image token and optional alias parse unambiguously.

## Test plan
- [x] Added `test_multistage_internal_from_no_warnings` covering a two-stage Dockerfile (`FROM node:20@sha256:abc123 AS builder` → `FROM builder AS runtime`)
- [x] `uv run --with pytest pytest tests/verify_action_build/test_security.py` — 16 passed
- [ ] Existing pinned/unpinned/tag-pinned/suspicious-curl cases still exercised

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -65,6 +65,18 @@ class TestAnalyzeDockerfile:
             warnings = analyze_dockerfile("org", "repo", "a" * 40)
         assert any("curl" in w.lower() or "evil" in w.lower() for w in warnings)
 
+    def test_multistage_internal_from_no_warnings(self):
+        dockerfile = (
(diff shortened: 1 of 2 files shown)


gopidesupavan (approved):
LGTM

PR #735, opened 2026-04-18
Remove matlab-actions/run-tests v3.1.0 from approved actions

potiuk:
Drop the 353aee49b0edf62278c118a51b484d90bf6da1b7 (v3.1.0) entry from actions.yml, approved_patterns.yml and the dependabot-review composite action. Earlier v2.3.1 and v3.0.0 hashes remain approved.

Generated-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


potiuk:
The matlab action 3.1.0 is "pre-release" and changes commit constantly.

potiuk:
We need to merge that one as it fails Zizmor checks as the commit is orphaned now - all other GH actions are failing because of it :(

dave2wave (approved):
Been following discussions and this makes sense

PR #740, opened 2026-04-19
Add stash uv.lock and enable uv exclude-newer cooldown

potiuk:
## Summary

- Add `stash/uv.lock` so the stash environment resolves to the same pinned set for contributors and CI.
- Enable `exclude-newer = "7 days"` under `[tool.uv]` in root, `pelican/`, and `stash/` pyproject.tomls, matching the existing Dependabot `cooldown.default-days: 7`.
- Register `/stash/` under the `uv` package-ecosystem in `.github/dependabot.yml` so Dependabot bumps `stash/uv.lock` on the same weekly cadence with the same cooldown.
- Bump `required-version = ">=0.9.17"` (the release that added friendly-duration support for `exclude-newer`).

## Why cooldowns + lockfile

Committed `uv.lock` files plus `exclude-newer` defend **both local development and CI** against supply-chain attacks that publish a malicious version of a dependency:

- **`uv.lock` pins exact versions and hashes.** Contributors running `uv sync` locally and CI workflows resolve to the same vetted set — nobody picks up a newer release by accident.
- **`exclude-newer = "7 days"` adds a release-age floor to every resolve.** When a lock *is* regenerated (manual bump, Dependabot, or a fresh `uv lock`), uv refuses to consider any package published in the last 7 days. A compromised upload has to sit public for a week before uv will resolve to it, giving the community time to detect, yank, or disclose before the bad version reaches anyone's machine.
- **Dependabot's own `cooldown.default-days: 7` covers the PR side.** The uv-level setting covers every *other* path — ad-hoc `uv lock`, a new contributor's first `uv sync`, CI jobs that do fresh resolves.

## Test plan

- [ ] CI green on this branch
- [ ] `uv sync` works in each of `/`, `pelican/`, `stash/` against a uv `>=0.9.17`
- [ ] Dependabot opens a PR for `stash/uv.lock` on the next weekly run


dave2wave (approved):
LGTM

Issue #686, opened 2026-04-09
Determine some more criteria for actions review


potiuk:
Following the discussion in https://github.com/apache/infrastructure-actions/pull/674#issuecomment-4216174449

Possible criteria:

- [x] Reproducibility — implemented in #561, #614, #629, #652, #674, #675, #734
- [x] Code review of differences vs. past approved version — implemented in #647, #671
- [x] Passing cooldown (now 0 days) — implemented in #560, #712, #737
- [x] Correctness of the action (for example errors in build pipelines) — ongoing fixes in #617, #628, #664, #685, #733, #734
- [x] Hash-pinning of composite actions — implemented in #629 (supporting fixes: #619, #620)
- [ ] Checking if downloaded binaries are verified — pending in #743
- [ ] Compatible licencing

Goal not criteria:

- [ ] Reject actions with suspicious builds, e.g. those that use different dependencies than what the source implies (proposed by @dave2wave) — partially addressed by verify-action-build warnings

Rejected:

~- [ ] Attestations~

Maybe others?

I would love to hear what others think.

potiuk:
I'd personally add [x] hash pinning and [x] Compatible licencing (even if optional and default is not-compatible). Possibly we should add some checks for the latter if the actions are used properly.

potiuk:
Also I think good idea when we reject some action we should follow up with the 3rd party - to at least let them know they should fix things (or even create PR).  That should be IMHO default behaviour when we reject things.

dave2wave:
#683 is for discussion of what cooldown periods should be recommended.

dave2wave:
What is the point of enforcing "compatible licensing"? Policy allows any build tool as long as it is unmodified.

dave2wave:
- [ ] Reject actions with suspicious builds like those that use different dependencies than what the source implies.

potiuk:
> * Reject actions with suspicious builds like those that use different dependencies than what the source implies.

I don't think this was "suspicious" - there is clear explanation on why it happened (and why it could have happened for others). The problem was (see my PR https://github.com/SonarSource/sonarqube-scan-action/pull/228 ) was that their dependabot PRs did not have a check if dist was rebuilt and release was made after that without rebuiliding it with the new tools.

[...]

Issue #665, opened 2026-04-04
Evaluate generation of aboms for ci/cd dependency analysis


dave2wave:
I saw a note about [Generating aboms](https://github.com/JulietSecurity/abom) on the OpenSSF slack. This tool may be a useful analysis gathering step in an atom analysis workflow.

Let's discuss.

dave2wave:
@potiuk  For python and npm type actions the verify-action-build script could be enhanced to provide a list of dependencies. These should be saved somewhere so that we can easily scan for exposure to vulnerable components.

potiuk:
> @potiuk For python and npm type actions the verify-action-build script could be enhanced to provide a list of dependencies. These should be saved somewhere so that we can easily scan for exposure to vulnerable components.

I don't think we need that - if anything, we should refuse those deps (and we already do in some cases) where "dependencies" in those actions are not fixed (I think vast majority - if not all of our actions have a "pinned" lock file (package.lock/ uv.lock) that contains all dependencies. 

So rather than  duplicating those we should reject those actions that do not have such a lock - because they are **dangerous** - as they can bring something we had no idea (and author of the action had no idea) at the time of release - this is what Astral does (I quoted their article recently) - ans this is what has been long requested from GitHub - they had "immutable" actions idea that they generally abandoned (for public actions). This has also been discussed by my friend - Andrew Nesbitt (from the OSS Corps of Security Enginers - https://nesbitt.io/2025/12/06/github-actions-package-manager.html  - and I quite agree - everything in our actions should be "locked/pinned".

potiuk:
Also to add: the reason why not in verify  - is because verify is "gating" tooll. Such a tool for abom storage should work **when** we merge somethign- not for everything. And it's a great oportunity for others to contribute.

So we could have a separate action/tool for it - maybe someone wouid like to use it as an opportunity to test their claude skills and add it. 

I added the customary "good first issue" label - so now the issue is in this nice https://github.com/apache/infrastructure-actions/contribute  page.

If we do not give people chance to contribute and try to piggy-back into existing (somewhat) related things and peopel who usually look at them we will never grow community.

potiuk:
BTW: checking for existing lock files for all possible ecosystems is in https://github.com/apache/infrastructure-actions/pull/770

PR #743, opened 2026-04-19
verify-action-build: add binary download verification check

potiuk:
## Summary

Implements the ["Checking if downloaded binaries are verified" criterion from #686](https://github.com/apache/infrastructure-actions/issues/686).

Adds a new `Binary Download Verification` check to `verify-action-build` that scans:
- Dockerfiles
- `action.yml` `run:` blocks
- Shell / Python / PowerShell scripts referenced by the action

For each file, it looks for binary or script downloads (curl, wget, Invoke-WebRequest, Dockerfile `ADD <url>`, pipe-to-shell) and requires a detectable verification step (sha*sum, shasum, md5sum, b2sum, openssl dgst, gpg --verify, cosign verify, slsa-verifier, gh attestation verify, minisign, ssh-keygen -Y verify, CertUtil -hashfile, Get-FileHash).

Package-manager invocations (apt/apk/pip/npm/brew/cargo/go install/etc.) are skipped because they provide their own integrity verification.

A file that contains downloads but no verification patterns causes the check to **fail** and the overall verification run to exit non-zero.

The check recurses through composite nested actions (MAX_DEPTH=3, skipping trusted orgs `actions` / `github` and unpinned refs), so a root composite action is held responsible for unverified downloads in the nested composites it calls.

Toggleable via `--no-binary-download-check` — enabled by default.

## Test plan

- [x] 14 new unit tests in `test_security.py` (135 tests total, all green):
  - pipe-to-shell fails
  - curl of `.tar.gz` without verification fails
  - sha256sum verification passes
  - gpg --verify passes
  - cosign verify-blob passes
  - `apk add` / `pip install` lines are not flagged
  - Dockerfile `ADD <url>` without verification fails
  - action.yml `run:` block scanning works (pass + fail cases)
  - GitHub releases binary path (`/releases/download/`) flagged
  - Recursion into a composite nested action surfaces its downloads
  - Nested `actions/checkout` (trusted org) is skipped
- [x] Test end-to-end against a real composite action with a known binary download
- [x] Test `--no-binary-download-check` flag disables the check

### DIFF
--- a/utils/verify_action_build/verification.py
+++ b/utils/verify_action_build/verification.py
@@ -192,6 +192,37 @@ def verify_single_action(
         checks_performed.append(("Action type detection", "info", action_type))
 
         is_js_action = action_type.startswith("node") or action_type in ("unknown",)
+
+        # Binary-download verification runs for every action type. JS actions
+        # can and do shell out to fetch pre-built binaries (e.g. platform-
+        # specific CLIs) from the action's own `run:` blocks or Dockerfile,
+        # and we want to flag those when they lack a checksum/signature step.
+        if check_binary_downloads:
+            bd_warnings, bd_failures = analyze_binary_downloads_recursive(
+                org, repo, commit_hash, sub_path,
+            )
+            binary_download_failures.extend(bd_failures)
+            if bd_failures:
+                checks_performed.append((
+                    "Binary download verification", "fail",
+                    f"{len(bd_failures)} unverified download(s)",
+                ))
+            elif bd_warnings:
+                checks_performed.append((
+                    "Binary download verification", "warn",
+                    f"{len(bd_warnings)} download(s); verification present",
+                ))
+            else:
+                checks_performed.append((
+                    "Binary download verification", "pass",
+                    "no downloads or all verified",
+                ))
+        else:
+            checks_performed.append((
+                "Binary download verification", "skip",
+                "disabled via --no-binary-download-check",
+            ))
+
         if not is_js_action:
             console.print()
             console.print(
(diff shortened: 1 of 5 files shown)


potiuk:
Pushed a follow-up: the binary-download check now runs for **every** action type (not just composite/docker) and additionally scans JS/TS sources for unverified downloads — catching patterns like `@actions/tool-cache`'s `tc.downloadTool` that the shell-only scanner missed entirely.

Re-ran against PR #724 (dependabot bump of `posit-dev/setup-air` 1.0.0 → 1.0.1), which had been passing this tool cleanly despite the action downloading a release tarball with no checksum verification. It now fails as it should:

```
───────────────────────── Binary Download Verification ─────────────────────────
  ✗ src/download/download-version.ts: 1 unverified JS download(s) (no
    checksum/signature check in file)
    line 39: const downloadPath = await tc.downloadTool(
```

```
───────────────────────────── Verification Summary ─────────────────────────────
                     posit-dev/setup-air@cf390573ff4f
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Check                          ┃ Status ┃ Detail                        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ Action type detection          │   ℹ    │ node24                        │
│ Binary download verification   │   ✗    │ 1 unverified download(s)      │
│ JS build verification          │   ✓    │ compiled JS matches rebuild   │
│ Approved versions              │   ℹ    │ 1 version(s) on file          │
│ Source diff vs approved        │   ℹ    │ compared against 63e80dedb6d2 │
└────────────────────────────────┴────────┴───────────────────────────────┘
```

Exit code 1 — previously the same run exited 0 because the scanner never looked inside `src/download/download-version.ts`.

Source patterns the JS/TS scanner now flags (with a required same-file verification step to count as "verified"):

- `tc.downloadTool(` / bare `downloadTool(` — `@actions/tool-cache`
- `fetch("https://…")`, node `http(s).get/request`, `axios.*`
- `new HttpClient(` — `@actions/http-client`
- `require("node-fetch")`

Verification counterparts: `crypto.createHash`, WebCrypto `subtle.digest`, sigstore/cosign references, `verifySignature`/`verifyChecksum`/`computeHash` helpers.

File discovery uses the GitHub trees API and scans the repo root plus `src/`, `lib/`, `source/`, `sources/`, `scripts/`. Explicitly excluded: `dist/`, `build/`, `out/`, `node_modules/`, `coverage/`, tests, examples, docs.

utils/verify_action_build/verification.py
```diff
@@ -270,6 +273,34 @@ def verify_single_action(
                 f"{len(repo_warnings)} warning(s)" if repo_warnings else "ok",
             ))
 
+            if check_binary_downloads:
```
dave2wave, line 276: An architectural note: it looks like each check does its own recursion to perform tests on the same objects. This design is completely understandable given the incremental way that this has been built.

dave2wave (approved):
Keep moving

utils/verify_action_build/verification.py
dave2wave: Specifically I'm thinking of license checking.

potiuk: Good point - we can ask claude to refactor it - as a followup (I do not like to mix features with refactors). It usually does a good job when you ask what you want.

potiuk: Opened #744 as a follow-up to consolidate the per-check recursion into a single walk — good call on raising this before the license-checking work lands.

dave2wave: Yes. I also have #665 in mind which really is the results of the downloads and licenses survey.

dave2wave: Maybe this should be a two step process?

PR #747, opened 2026-04-19
verify-action-build: fix RESULT panel cause on binary-download failure

potiuk:
## Summary

Observed on [PR #724's verify run](https://github.com/apache/infrastructure-actions/actions/runs/24634855262/job/72028382958?pr=724): the summary table correctly showed `JS build verification: ✓ compiled JS matches rebuild` and `Binary download verification: ✗ 1 unverified download(s)` — but the RESULT panel still said "Differences detected between published and rebuilt JS", which was not the cause of failure.

## Root cause

In `verify_single_action`, the failure branch chose its message based on `is_js_action` first and hardcoded the JS-mismatch string for every JS action. Since #743, the binary-download check runs for *every* action type (including JS actions), so `overall_passed = all_match and not binary_download_failures` can go false via the second term alone — but the RESULT text would still blame JS.

## Fix

Dispatch on the actual failure condition:
- `all_match == False` → JS-mismatch message (with or without the lock-file-retry variant).
- Otherwise, `binary_download_failures` → binary-download message.
- Fallback → generic "verification failed".

## Test plan

- [x] New `TestVerifySingleActionResultMessage` class in `test_verification.py`:
  - `test_js_action_with_unverified_download_reports_binary_cause` — the regression case; **fails without the fix**, passes with it
  - `test_js_action_with_js_mismatch_reports_js_cause` — guards the original JS-mismatch path stays intact
- [x] Full `tests/verify_action_build/` suite green: 137 passed (was 135 before the 2 new tests)
- [x] Re-run verify on PR #724 after merge to confirm the RESULT panel reads correctly

## Related

- #743 — introduced the binary-download check
- PR #724 — run that surfaced the incorrect message

### DIFF
--- a/utils/tests/verify_action_build/test_verification.py
+++ b/utils/tests/verify_action_build/test_verification.py
@@ -254,3 +254,108 @@ class TestVerifySingleActionLockFileRetry:
         assert len(js_rows) == 1
         assert js_rows[0][1] == "fail"
         assert "approved lock files" in js_rows[0][2]
+
+
+class TestVerifySingleActionResultMessage:
+    """The RESULT panel text must describe the actual failure cause."""
+
+    def _patch_stack(
+        self,
+        *,
+        approved: list[dict],
+        diff_js_side_effect,
+        binary_download_result,
+        action_type: str = "node20",
+    ):
+        patches = {
+            "parse_action_ref": mock.patch(
+                "verify_action_build.verification.parse_action_ref",
+                return_value=("org", "repo", "", "c" * 40),
+            ),
+            "find_approved_versions": mock.patch(
+                "verify_action_build.verification.find_approved_versions",
+                return_value=approved,
+            ),
+            "build_in_docker": mock.patch(
+                "verify_action_build.verification.build_in_docker",
+                return_value=_build_in_docker_result(action_type=action_type),
+            ),
+            "diff_js_files": mock.patch(
+                "verify_action_build.verification.diff_js_files",
+                side_effect=diff_js_side_effect,
+            ),
+            "show_approved_versions": mock.patch(
+                "verify_action_build.verification.show_approved_versions",
+                return_value=None,
+            ),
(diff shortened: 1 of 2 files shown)


potiuk:
Verified end-to-end on `posit-dev/setup-air@cf390573ff4fe0f198f35df3a642b1409328d859` (the ref bumped by PR #724) using this branch:

```
Verification Summary
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Check                          ┃ Status ┃ Detail                        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ Action type detection          │   ℹ    │ node24                        │
│ Binary download verification   │   ✗    │ 1 unverified download(s)      │
│ JS build verification          │   ✓    │ compiled JS matches rebuild   │
│ Approved versions              │   ℹ    │ 1 version(s) on file          │
│ Source diff vs approved        │   ℹ    │ compared against 63e80dedb6d2 │
└────────────────────────────────┴────────┴───────────────────────────────┘

╭─────────────────────────────────── RESULT ───────────────────────────────────╮
│ node24 action — 1 unverified binary download(s) detected (no                 │
│ checksum/signature check in file)                                            │
╰──────────────────────────────────────────────────────────────────────────────╯
Exit code: 1
```

The RESULT panel now correctly attributes the failure to the binary-download check (`tc.downloadTool` in `src/download/download-version.ts`) instead of misleadingly saying "Differences detected between published and rebuilt JS" as in the original [PR #724 run](https://github.com/apache/infrastructure-actions/actions/runs/24634855262/job/72028382958?pr=724). Exit code 1 as expected.

raboof (approved):

PR #751, opened 2026-04-20
Allow erlef/setup-beam v1.24.0

fresh-borzoni:

## Overview
closes https://issues.apache.org/jira/browse/INFRA-27826

Allow `apache/fluss-rust` (incubating) to use `erlef/setup-beam` for CI of its Elixir bindings. `erlef/setup-beam` is the de-facto standard setup action for BEAM-language projects (Erlang, Elixir, Gleam) and is maintained by the Erlang Ecosystem Foundation — a vendor-neutral organization supporting the Erlang/Elixir ecosystem. It is the direct equivalent of `actions/setup-python` or `actions/setup-node` for the BEAM VM.

Context:
- INFRA ticket: https://issues.apache.org/jira/browse/INFRA-27826
- Consumer PR: https://github.com/apache/fluss-rust/pull/497

**Name of action:** `erlef/setup-beam`

**URL of action:** https://github.com/erlef/setup-beam

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):** `fc68ffb90438ef2936bbb3251622353b3dcb2f93` (v1.24.0)

## Permissions

The action installs Erlang/OTP and Elixir (or other BEAM toolchains) onto the runner. It requires no elevated repository permissions and does not access credentials. Calling workflows in apache/fluss-rust will keep `permissions:` minimal (default `contents: read`).

## Related Actions

No existing allowlisted action covers BEAM/Elixir toolchain setup. The closest analogues are `actions/setup-python`, `actions/setup-node`, and `astral-sh/setup-uv`, which handle other language ecosystems.

## Checklist

- [x] The action has a clearly defined license (MIT)
- [x] The action has CI/unit tests configured (ubuntu.yml, macos.yml, windows.yml)
- [x] Compiled JavaScript in `dist/` matches a clean rebuild (verified with `uv run utils/verify-action-build.py erlef/setup-beam@fc68ffb90438ef2936bbb3251622353b3dcb2f93` — all compiled JS matches, no binary downloads, action type node24)

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -376,6 +376,9 @@ erisu/license-checker-action:
     expires_at: 2026-07-05
   04511f4c052b5773f11e1c65b42cda88235c62ae:
     tag: v2.1.0
+erlef/setup-beam:
+  fc68ffb90438ef2936bbb3251622353b3dcb2f93:
+    tag: v1.24.0
 geekyeggo/delete-artifact:
   '*':
     keep: true


fresh-borzoni:
@potiuk 
Q: CI flags src/setup-beam.js:1241 - tc.downloadTool over HTTPS to GitHub releases / hexpm, same pattern as actions/setup-python, actions/setup-go, ruby/setup-ruby, dart-lang/setup-dart. 

If I read it correctly -the check looks for any crypto-verification token  anywhere in the same file, which setup-go/setup-ruby pass on an incidental crypto.createHash for cache-key naming, for example.
                                                                                                                                                                                                                                                     Should this block approval, or does the language-toolchain-setup case sit alongside the apt/pip whitelist rationale?

potiuk:
> @potiuk Q: CI flags src/setup-beam.js:1241 - tc.downloadTool over HTTPS to GitHub releases / hexpm, same pattern as actions/setup-python, actions/setup-go, ruby/setup-ruby, dart-lang/setup-dart.
> 
> If I read it correctly -the check looks for any crypto-verification token anywhere in the same file, which setup-go/setup-ruby pass on an incidental crypto.createHash for cache-key naming, for example. Should this block approval, or does the language-toolchain-setup case sit alongside the apt/pip whitelist rationale?


That's a good question - also to @dave2wave @raboof and @ppkarwasz and something we discuss in https://github.com/apache/infrastructure-actions/issues/686 - what should be criteria for verifying actions. @ppkarwasz came with the idea of verifying it - and I implemented it literally yesterday ...

And I think we have not decided yet. WDYT?  @dave2wave @raboof @ppkarwasz.

IMHO - this is the same case as with Correctness of the action - we should still have the check and decide individually - while raise a PR to an action creator to fix this in the next version. @fresh-borzoni - would you be willing to raise such PR and follow up with the action creators if we let it pass this time ?  I think it would be a good contribution to the process as well :)

fresh-borzoni:
Thanks @potiuk, doing it. 
 
Opened an issue at erlef/setup-beam proposing 2 options to address this: https://github.com/erlef/setup-beam/issues/456
Once the prefered direction is clear, I can help with the PR.

ppkarwasz (approved):
The Erlang Ecosystem Foundation is very security-conscious, so I've reached out to them for guidance on the best way to cryptographically secure the supply chain from OTP through to `erlef/setup-beam`. I expect feedback in the next couple of days.

The setup here is quite similar to `aquasecurity/setup-trivy`, so it's worth turning this into a good example of cross-project collaboration. A few observations on the current state:

- Most commits in `erlef/setup-beam` are signed, but neither the release tag nor the commit it points to is signed. There's room for improvement here: if tags were signed, we could hardcode the release managers' GPG keys and verify against them.
- The action downloads artifacts without verifying them, which is the problem tracked in erlef/setup-beam#456.

However, this setup isn't materially worse than `actions/setup-java` and similar actions, so I'm not opposed to merging now and tracking the hardening work separately.

potiuk:
I also have no problem with merging especially that we  have an issue tracking it - if not addressed we can make different decision for next upgrade

potiuk (approved):

potiuk:
Check looks good besides the binary verification (the error message is misleading and already fixed in #747

potiuk:
@raboof @dave2wave ?

dave2wave:
OFC: Is there a trademark issue between this BEAM and Apache Beam?

potiuk:
> OFC: Is there a trademark issue between this BEAM and Apache Beam?

I guess you will haave to check with Beam PMC, but I highly doubt, beam is quite a bit too common word to make a successful claim there.

potiuk:
@dependabot rebase

potiuk:
both me and @ppkarwasz are ok - so I will merge for now to not hold @fresh-borzoni any more. Thanks @fresh-borzoni for raising the issue - I see things are progressing, so hopefully next version will be doing verification, if not we wil re-flag it to them

potiuk (approved):

fresh-borzoni:
@potiuk @ppkarwasz Thank you 🙇

potiuk:
> @potiuk @ppkarwasz Thank you 🙇
You are most welcome!

PR #756, opened 2026-04-21
Add pip-audit action with version and expiration

tisonkun:
This works around https://github.com/apache/infrastructure-actions/pull/452#issuecomment-4286855602.

Not sure if there are any better options.

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -744,10 +744,6 @@ pre-commit/action:
 PyO3/maturin-action:
   '*':
     keep: true
-pypa/gh-action-pip-audit:
-  1220774d901786e6f652ae159f7b6bc8fea6d266:
-    tag: v1.1.0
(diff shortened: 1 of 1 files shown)


actions.yml
```diff
@@ -744,6 +744,10 @@ pre-commit/action:
 PyO3/maturin-action:
   '*':
     keep: true
+pypa/gh-action-pip-audit:
+  1220774d901786e6f652ae159f7b6bc8fea6d266:
+    tag: v1.1.0
+    expires_at: 2026-07-31
```
raboof: you don't need to add an expiration, an expiration will be automatically added when the next version is allowlisted

tisonkun: Thanks! Removed.

raboof (approved):
verified this re-adds the previously-reviewed action

PR #768, opened 2026-04-24
verify-action-build: handle source-detached orphan release tags

potiuk:
## Summary

- Detect release tags whose tree is a rootless orphan commit containing only `action.yml`, `dist/`, `LICENSE`, `README.md` — no `src/` and no `package.json` to rebuild from. Common with changesets-driven release flows (e.g. slackapi/slack-github-action v3.0.x).
- Resolve the corresponding default-branch source commit via the GitHub Releases API: match tag → `published_at` → most recent default-branch commit at or just before `published_at` with a buildable `package.json`, preferring "chore: release"-style messages.
- Docker build now captures `/original-dist` from the orphan tag, then `git checkout`s to the resolved source commit before building, so the rebuild runs against real source and the diff compares against the tag's published `dist/`.

## Why

Currently verifier runs on `slackapi/slack-github-action@v3.0.2` (PR #762) fail with "JS build verification: DIFFERENCES DETECTED" — but the actual cause is that the tag commit has no source at all, so the rebuild silently produces nothing. The failure message is misleading and blocks auto-approval of an otherwise-valid bump. Verified end-to-end:

- `slackapi/slack-github-action@v3.0.2` (`03ea5433c137`) → resolves to `90fee45e5a24` (main's "chore: release (#593)"), rebuilt `dist/index.js` + `dist/sourcemap-register.cjs` are **byte-identical** to the tag's published artifacts.
- `v3.0.1` (same pattern, already approved) also passes cleanly under the new path.
- `actions/checkout`, `astral-sh/setup-uv`, `scacap/action-surefire-report` are **not** flagged as detached — detection is narrow (dist/ present, package.json + src/ absent, top-level only), so the normal build path is unchanged.

## Test plan

- [ ] Re-run CI on #762 and confirm `verify` now passes.
- [ ] Spot-check the new path against another changesets-style action (if any) in a future dependabot PR.
- [ ] Regression check: existing dependabot PRs with normal (non-detached) tags continue to verify as before.

### DIFF
--- a/utils/verify_action_build/docker_build.py
+++ b/utils/verify_action_build/docker_build.py
@@ -104,12 +104,18 @@ def build_in_docker(
     cache: bool = True,
     show_build_steps: bool = False,
     approved_hash: str = "",
+    source_commit_hash: str = "",
 ) -> tuple[Path, Path, str, str, bool, Path, Path]:
     """Build the action in a Docker container and extract original + rebuilt dist.
 
     When *approved_hash* is supplied the Docker build restores package lock files
     from that commit so the rebuild uses the same dev-dependency versions that
     produced the original dist/.
+
+    When *source_commit_hash* is supplied the Docker build captures the original
+    dist/ from *commit_hash* (a source-detached release tag) and then switches
+    the tree to *source_commit_hash* before building.  Used for actions whose
+    tagged commit is an orphan tree without buildable source.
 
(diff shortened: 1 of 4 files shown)


potiuk:
This is apparently common practice that action is released from an Orphan tag that "only" contains compiled javascript - no sources. This change to our verify script should handle it.

dave2wave (approved):
Looks like a necessary new feature

PR #775, opened 2026-04-25
verify-action-build: skip binary-download check for pure data fetches

potiuk:
## Summary

The binary-download check currently flags every `tc.downloadTool` / `fetch` / `https.get` in JS source that doesn't sit alongside a hash/signature pattern. That over-flags actions that fetch HTTP responses **as data** — a metadata badge regex'd for an integer, an API JSON parsed for a version number — where there is no binary or executable to verify in the first place.

Add a per-file heuristic that exempts pure data fetches from the check.

## The false-positive that motivated this

PR #752 (`dependabot/fetch-metadata` 3.0.0 → 3.1.0) was flagged on `src/dependabot/verified_commits.ts:99`:

```ts
const svg = await new Promise<string>((resolve) => {
  https.get(`https://dependabot-badges.githubapp.com/badges/compatibility_score?...`, res => { ... })
})
const scoreChunk = svg.match(/<title>compatibility: (?<score>\d+)%<\/title>/m)
return scoreChunk?.groups ? parseInt(scoreChunk.groups.score) : 0
```

This is a computed-on-demand SVG badge — no possible "expected hash" exists, and a regex-extracted integer doesn't cross a security boundary. Adding a checksum upstream would be cargo-cult.

## The heuristic

A file's download calls are exempted iff **both** hold:

1. The file contains at least one **data-parse marker**: `.match` / `.matchAll` / `.replace` / `.test` / `.split` / `.includes` / `.startsWith` / `.endsWith` / `.toLowerCase()` / `.toUpperCase()` / `JSON.parse` / `parseInt` / `parseFloat` / `Number(...)`.
2. The file contains **no binary-handling markers**: `tc.extractTar` / `tc.extractZip` / `tc.extract7z` / `tc.extractXar`, `tc.cacheFile` / `tc.cacheDir`, `fs.writeFile` family, `exec.exec` / `child_process` / `spawn`, `chmod` calls, `'chmod +x'` strings.

Per-file scope (not per-finding): a file that mixes a metadata fetch with a real archive download keeps the strict check on **every** download in it. Per-finding scoping would need a real AST to track which response gets passed to which use, and the conservative blanket rule is preferable to a flaky one.

## Verified behaviour

| Action | Before | After |
|---|---|---|
| `dependabot/fetch-metadata@v3.1.0` (the false positive) | ✗ 1 unverified JS download | ✓ no downloads or all verified |
| `leafo/gh-actions-luarocks@v6.1.0` (real `tc.downloadTool` + `tc.extractTar`) | ✗ 1 unverified JS download | ✗ 1 unverified JS download (still flagged — correct) |

## Test plan

- [x] 7 new unit tests in `test_security.py::TestPureDataFetchExemption` covering: pure-data-fetch exempted, real-binary still flagged, mixed-file not exempt, no-data-parse-marker not exempt, `chmod +x` string disables exemption, `JSON.parse` alone exempts, `.split` alone exempts.
- [x] Full suite: 146 tests pass.
- [x] Live smoke against `dependabot/fetch-metadata@25dd0e34` — `Binary download verification: ✓` (was ✗).
- [x] Live smoke against `leafo/gh-actions-luarocks@35d062de` — still ✗ (the real flag we want).

### DIFF
--- a/utils/verify_action_build/security.py
+++ b/utils/verify_action_build/security.py
@@ -1056,6 +1056,43 @@ _JS_DOWNLOAD_PATTERNS = [
     re.compile(r"\brequire\(\s*['\"`]node-fetch['\"`]"),
 ]
 
+# Markers indicating the response of an HTTP call is consumed as *data* —
+# parsed for a value rather than written to disk and executed/extracted.
+# A "data fetch" with these markers and no binary-handling markers (see
+# below) anywhere in the file is exempt from the unverified-download check,
+# since there is no binary or executable to verify in the first place.
+_JS_DATA_PARSE_PATTERNS = [
+    re.compile(r"\.match\s*\("),
+    re.compile(r"\.matchAll\s*\("),
+    re.compile(r"\.replace\s*\("),
+    re.compile(r"\.test\s*\("),
+    re.compile(r"\bJSON\.parse\s*\("),
+    re.compile(r"\bparseInt\s*\("),
+    re.compile(r"\bparseFloat\s*\("),
+    re.compile(r"\bNumber\s*\("),
+    re.compile(r"\.split\s*\("),
+    re.compile(r"\.includes\s*\("),
+    re.compile(r"\.startsWith\s*\("),
+    re.compile(r"\.endsWith\s*\("),
+    re.compile(r"\.toLowerCase\s*\(\)"),
+    re.compile(r"\.toUpperCase\s*\(\)"),
+]
+
+# Markers indicating the response is treated as a binary or executable —
+# extracted, written to disk, made executable, or executed.  If any of these
+# appear anywhere in the file the download stays subject to the verification
+# requirement, regardless of any data-parse markers also present.
+_JS_BINARY_HANDLE_PATTERNS = [
+    re.compile(r"\btc\.extract(?:Tar|Zip|7z|Xar)\s*\("),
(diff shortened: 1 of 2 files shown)


utils/verify_action_build/security.py
```diff
@@ -853,6 +853,43 @@ def analyze_action_metadata(
...
+    re.compile(r"\bparseInt\s*\("),
+    re.compile(r"\bparseFloat\s*\("),
+    re.compile(r"\bNumber\s*\("),
+    re.compile(r"\.split\s*\("),
+    re.compile(r"\.includes\s*\("),
+    re.compile(r"\.startsWith\s*\("),
+    re.compile(r"\.endsWith\s*\("),
+    re.compile(r"\.toLowerCase\s*\(\)"),
+    re.compile(r"\.toUpperCase\s*\(\)"),
+]
```
dave2wave, line 1079: How was this list determined?

potiuk: By asking Claude. It saw enough of a code to be able to determine it statistically.

potiuk:
rebased

utils/verify_action_build/security.py
potiuk: The complete heuristics is explained in the PR description

dave2wave (approved):
LGTM

PR #777, opened 2026-04-26
verify-action-build: keep non-minified JS, diff vs approved version

potiuk:
## Summary
- For non-minified compiled JS (Deno bundle output, Dart `dart compile js`, esbuild readable output) the rebuild produces toolchain-version noise rather than actionable diffs — surfaced as "non-minified — rebuild differs" warnings on PRs like #736.
- Skip those files in the Dockerfile pre-rebuild deletion step (record paths in `/kept-js.log`) and diff them against the previously-approved version instead, with `beautify_js()` applied so the diff renders readably.
- Minified bundles still take the full delete-and-rebuild path, so hand-written or unbuilt JS in `dist/` is still caught.

## Test plan
- [ ] `uv run pytest` — 199 tests pass locally; new `test_keeps_non_minified_compiled_js` and `TestDiffJsKeptFiles` cover the Dockerfile heuristic and the diff_js short-circuit
- [ ] Re-run the `verify` workflow on PR #736 and confirm `post.js` is reported as kept (yellow `~`, "diffed against previously-approved version below") rather than "rebuild differs"
- [ ] Confirm `dist/post.js` appears in the approved-vs-new diff section, beautified

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,8 @@ The script will:
 
 A clean result confirms that the compiled JS was built from the declared source. Any differences will be flagged for manual inspection.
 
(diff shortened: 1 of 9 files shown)


raboof (approved):
sounds reasonable

PR #781, opened 2026-04-28
check-for-transitive-failures: hourly cron with change-aware gate

potiuk:
## Summary

The `check-for-transitive-failures` workflow has been failing on virtually every Dependabot merge — see [run 25040287968](https://github.com/apache/infrastructure-actions/actions/runs/25040287968) and [recent history](https://github.com/apache/infrastructure-actions/actions/workflows/check-for-transitive-failures.yml). Daily cron runs always pass; push-triggered runs after Dependabot merges almost always fail with "action ... is not allowed".

Root cause: GitHub validates every action SHA in the composite against the **org-level** allowlist, which is synced from `approved_patterns.yml` by an external ASF Infra job. When a Dependabot PR merges, three things race:
1. The check-for-transitive-failures workflow fires on the push.
2. `update_actions.yml` regenerates `actions.yml` + `approved_patterns.yml`.
3. The external ASF sync picks up the regenerated `approved_patterns.yml` and updates the org allowlist.

The check loses the race almost every time. PR #721 (mine) tried to fix this by scoping `push` to `main`, on the assumption the regenerator would land first — but both workflows fire concurrently on the same merge commit and the external sync is asynchronous to either.

## Approach

- Drop the `push` trigger.
- Run **hourly** instead of daily.
- Add a cheap **gate** job that decides whether the real check needs to run by diffing the composite `action.yml` against the SHA of the last successful run.
- **30-min grace period**: skip if the composite changed less than 30 min ago, giving the external sync time to land.
- **24-hour safety net**: always run if it's been a day, preserving the original "detect a transitive dep that fell out of the allowlist without any local edit" purpose.

Most hours: gate runs in ~5 seconds, decides nothing changed, exits. After a Dependabot merge: gate skips for the first hour (grace), then triggers the actual check on the next hourly tick — by which point the org sync should have landed.

## Trade-offs

- The 30-min grace period is a guess; if ASF Infra's sync sometimes takes longer we'll still see occasional fail-then-pass cycles. Bumping `GRACE_MINUTES` is a one-line tune.
- Detection latency for genuine transitive drop-outs is now ≤1h (down from ≤24h) for changed-action cases, and remains ≤24h for the no-local-change safety net.

## Test plan

- [ ] Merge, then watch the next hourly run skip (gate exits with `should_run=false` since composite unchanged).
- [ ] On the next Dependabot merge, watch: hour-of-merge skips (grace), hour-after merges and the check passes.
- [ ] Manually trigger via `workflow_dispatch` to confirm the check still works.

### DIFF
--- a/.github/workflows/check-for-transitive-failures.yml
+++ b/.github/workflows/check-for-transitive-failures.yml
@@ -19,22 +19,119 @@
 
 name: Check for transitive failures in current latest actions
 
+# Runs hourly. The `gate` job decides whether the actual `check` job
+# needs to run, so most hours we burn ~5 seconds on the gate and skip
+# the rest. The `check` job is what actually exercises the org-wide
+# allowlist by loading the composite action.
+#
+# Triggering on `push` after a Dependabot merge does not work: the
+# org-level allowlist is synced from approved_patterns.yml by an
+# external ASF Infra job that runs on its own cadence, so the new SHA
+# is typically not yet on the allowlist when the push fires. The
+# 30-minute grace period below gives that sync time to land before we
+# probe.
 on:
   workflow_dispatch:
   schedule:
-    # Run daily to detect when a transitive dependency has fallen out
-    # of the allowlist even without changes to the action file.
-    - cron: "30 5 * * *"
-  push:
-    branches:
-      - main
-    paths:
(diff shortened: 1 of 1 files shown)


dave2wave (approved):
LGTM

PR #789, opened 2026-05-01
Add rubygems/release-gem to allowlist

kpumuk:
# Request for adding a new GitHub Action to the allow list

## Overview

Add `rubygems/release-gem` to the allowlist. This is the official trusted publishing action maintained by the RubyGems team, see [Releasing gems with a trusted publisher](https://guides.rubygems.org/trusted-publishing/#releasing-gems-with-a-trusted-publisher).

At Apache Thrift we [want to switch to trusted publishing](https://github.com/apache/thrift/pull/3383) to remediate supply chain risk, but I believe the workflow will be blocked until the action is allowlisted. See https://github.com/apache/thrift/blob/master/.github/workflows/release_ruby.yml

**Name of action:**

rubygems/release-gem

**URL of action:**

https://github.com/rubygems/release-gem

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):**

`6317d8d1f7e28c24d28f6eff169ea854948bd9f7`

## Permissions

`id-token: write`

## Related Actions

* `rubygems/configure-rubygems-credentials@bc6dd217f8a4f919d6835fcfefd470ef821f5c44 # v1`
   https://github.com/rubygems/release-gem/blob/v1/action.yml#L58C13-L58C93

## Checklist
You should be able to check most of these boxes for an action to be considered for review.
Please check all boxes that currently apply:

- [x] The action is listed in the GitHub Actions Marketplace
- [x] The action is not already on the list of approved actions
- [x] The action has a sufficient number of contributors or has contributors within the ASF community
- [x] The action has a clearly defined license
- [x] The action is actively developed or maintained
- [x] The action has CI/unit tests configured
- [x] Compiled JavaScript in `dist/` matches a clean rebuild (verify with `uv run utils/verify-action-build.py org/repo@hash`)

PR #790, opened 2026-05-02
verify-action-build: recognize @actions/http-client JSON helpers as data fetches

potiuk:
## Summary
- The `verify` check on #789 (rubygems/release-gem) failed on a false positive: it transitively scanned `rubygems/configure-rubygems-credentials@bc6dd217...` (composite action → node20 helper) and flagged two OIDC token-exchange files that POST a JWT to `${server}/api/v1/oidc/...` via `@actions/http-client`'s `postJson`. The response there is a short-lived credential validated through a Zod schema — not a binary or executable.
- Extends the existing `_file_is_pure_data_fetch` heuristic (#775) by recognizing the `*Json` family (`postJson` / `getJson` / `putJson` / `patchJson` / `delJson` / `requestJson`) as data-parse markers. These methods auto-parse the response body as JSON, so reaching for them is an explicit "consume as structured data" signal.
- The binary-handle gate is unchanged: if the same file also calls `tc.extractTar` / `fs.writeFile` / `exec.exec` / `chmod +x` / etc., the exemption still does not apply.

## Test plan
- [x] `pytest tests/verify_action_build/test_security.py -q` — 78/78 pass, including 3 new tests:
  - `test_postJson_token_exchange_exempted` — mirrors the exact PR #789 shape (HttpClient.postJson → OIDC exchange).
  - `test_getJson_alone_exempts` — sibling getJson.
  - `test_postJson_with_extract_in_same_file_not_exempt` — guard: a `getJson` next to `tc.downloadTool` + `tc.extractTar` keeps both flagged.
- [ ] After merge, re-run the verify check on #789 and confirm it now passes.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -863,3 +863,55 @@ return lines[0]
 """
         assert _file_is_pure_data_fetch(split_fetch) is True
         assert _find_binary_downloads_js(split_fetch) == []
+
+    # The exact pattern that triggered the false positive on PR #789
+    # (rubygems/configure-rubygems-credentials, transitively pulled in by
+    # rubygems/release-gem): @actions/http-client postJson against an OIDC
+    # token-exchange endpoint. The response is a credential, not a binary.
+    RUBYGEMS_OIDC_EXCHANGE = """\
+import * as core from '@actions/core'
+import {HttpClient} from '@actions/http-client'
+import {IdToken, IdTokenSchema} from './responses'
+
+export async function exchangeToken(audience, server) {
+  const webIdentityToken = await core.getIDToken(audience)
+  const http = new HttpClient('rubygems-oidc-action')
+  const url = `${server}/api/v1/oidc/trusted_publisher/exchange_token`
+  const res = await http.postJson(url, {jwt: webIdentityToken}, {})
+  return res.result
+}
+"""
+
+    def test_postJson_token_exchange_exempted(self):
+        # @actions/http-client postJson is JSON-only — response is parsed as
+        # structured data, never persisted or executed.
+        assert _file_is_pure_data_fetch(self.RUBYGEMS_OIDC_EXCHANGE) is True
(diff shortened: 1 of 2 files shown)


kpumuk (approved):

utils/tests/verify_action_build/test_security.py
```diff
@@ -863,3 +863,55 @@ def test_split_alone_exempts(self):
...
+import {HttpClient} from '@actions/http-client'
+import {IdToken, IdTokenSchema} from './responses'
+
+export async function exchangeToken(audience, server) {
+  const webIdentityToken = await core.getIDToken(audience)
+  const http = new HttpClient('rubygems-oidc-action')
+  const url = `${server}/api/v1/oidc/trusted_publisher/exchange_token`
+  const res = await http.postJson(url, {jwt: webIdentityToken}, {})
+  return res.result
+}
```
dave2wave, line 882: Is the plan to handle each of the OIDC / Trusted publishing actions as they upgrade and fail?

Also, it would be good to think about tagging these as it is important to identify all of the Trusted publishing actions that PMCs are using.

utils/verify_action_build/security.py
```diff
@@ -1076,6 +1076,11 @@ def analyze_action_metadata(
     re.compile(r"\.endsWith\s*\("),
     re.compile(r"\.toLowerCase\s*\(\)"),
     re.compile(r"\.toUpperCase\s*\(\)"),
+    # @actions/http-client JSON helpers (postJson/getJson/putJson/patchJson/
+    # delJson/requestJson) auto-parse the response body as JSON. Reaching for
+    # these is an explicit "treat the response as structured data" signal
+    # — typical of OIDC/RPC token-exchange calls, not binary downloads.
+    re.compile(r"\.(?:get|post|put|patch|del|request)Json\s*\("),
```
dave2wave, line 1083: This handles other generic OIDC token exchanges?

dave2wave (approved):
Small incremental changes

PR #798, opened 2026-05-04
verify-action-build: match TS generics on @actions/http-client *Json calls

potiuk:
## Summary

- The earlier fix (commit 920d616) added `*Json` to the data-parse markers but the regex required `(` directly after `Json`, so it silently missed the real `http.postJson<IdToken>(...)` call in rubygems/configure-rubygems-credentials v2.0.0 — PR #795 still flagged both `src/oidc/assumeRole.ts` and `src/oidc/trustedPublisher.ts` as unverified downloads.
- Allow an optional `<...>` generic between `Json` and `(`.
- Rewrite the `RUBYGEMS_OIDC_EXCHANGE` test fixture so it mirrors the v2.0.0 source verbatim (with the `<IdToken>` generic and `IdTokenSchema.parse(res.result)`) — closing the regression-test hole that let this slip through the first time.

## Test plan

- [x] `uv run pytest utils/tests/verify_action_build/test_security.py -k 'PureDataFetch or postJson or getJson'` — 10/10 pass.
- [x] Spot-checked against the real source at PR #795's SHA: `_file_is_pure_data_fetch` is `True` and `_find_binary_downloads_js` is `[]` for both flagged files.
- [ ] Re-run the "Handle Dependabot update" check on PR #795 once this lands on main — expect no unverified-download findings.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -864,10 +864,12 @@ return lines[0]
         assert _file_is_pure_data_fetch(split_fetch) is True
         assert _find_binary_downloads_js(split_fetch) == []
 
-    # The exact pattern that triggered the false positive on PR #789
+    # The exact pattern that triggered the false positive on PR #789 / #795
     # (rubygems/configure-rubygems-credentials, transitively pulled in by
     # rubygems/release-gem): @actions/http-client postJson against an OIDC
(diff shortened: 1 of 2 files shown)


potiuk:
ANther false-positive to handle :)

dave2wave (approved):

PR #800, opened 2026-05-04
verify-action-build: recognize sibling verification steps

potiuk:
## Summary
Two complementary changes that together let `sbt/setup-sbt@v1.1.23` pass binary-download verification:

1. **Recognize sibling verification steps** — within a single composite `action.yml`, treat verification as an action-wide property: a sibling step that does shell-level verification (`sha256sum`, `gpg --verify`, …) **or** invokes a known verifier (`carabiner-dev/actions/ampel/verify`, `slsa-framework/slsa-verifier-action`, `actions/attest-verify`) covers downloads in any of the action's run blocks or referenced scripts.

2. **Treat verifier-action refs as trust leaves** — when the recursive walker encounters a hash-pinned ref to a known verifier action, it stops there (same handling as `TRUSTED_ORGS`). Otherwise the walker would surface the verifier's own bootstrap installer (e.g. `carabiner-dev/actions/install/ampel`), which can't verify the verifier binary it installs (chicken-and-egg). The trust anchor is the hash-pinned commit of the verifier ref.

Fixes the false-positive flagged on #799 for `sbt/setup-sbt@v1.1.23` (the action downloads `.zip`+`.zip.asc` in one step and verifies the signature in a sibling Ampel step).

## Test plan
- [x] All 203 existing `verify_action_build` tests pass.
- [x] 4 new tests added: Ampel sibling step passes, SLSA verifier sibling passes, unrelated `uses:` step still fails, recursive walker stops at a verifier ref.
- [x] End-to-end verified against `sbt/setup-sbt@v1.1.23` via `uv run utils/verify-action-build.py`: Binary download verification now reports ✓ instead of ✗.

PR #802, opened 2026-05-05
Add carabiner-dev/actions/ampel/verify action (v1.2.0)

pjfanning:
Apache Pekko builds that use setup-sbt v1.1.23 run into https://github.com/sbt/setup-sbt/issues/84

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -165,6 +165,9 @@ burrunan/gradle-cache-action:
 bytedeco/javacpp-presets:
   '*':
     keep: true
+carabiner-dev/actions/ampel/verify:
+  360ffa1eb909b0105d4eccb6d6ef337911c34952:
+    tag: v1.1.6
 carloscastrojumo/github-cherry-pick-action:
   503773289f4a459069c832dc628826685b75b4b3:
     tag: v1.0.10
--- a/actions.yml
+++ b/actions.yml
@@ -166,8 +166,8 @@ bytedeco/javacpp-presets:
   '*':
     keep: true
 carabiner-dev/actions/ampel/verify:
-  360ffa1eb909b0105d4eccb6d6ef337911c34952:
-    tag: v1.1.6
+  e0e3b8149dafed833431095bc148d50e7eade4e8:
+    tag: v1.2.0
 carloscastrojumo/github-cherry-pick-action:
   503773289f4a459069c832dc628826685b75b4b3:
     tag: v1.0.10


raboof:
The 'verify' build is failing with

```
───────────────────────────────────────────────────────────────────────────────────── Binary Download Verification ─────────────────────────────────────────────────────────────────────────────────────
  ✓ No binary downloads detected

───────────────────────────────────────────────────────────────────────────────────── Binary Download Verification ─────────────────────────────────────────────────────────────────────────────────────
  ✗ action.yml : 1 unverified download(s) (no checksum/signature check in file)
    line 2: curl -Lo ${{ inputs.install-dir }}/bin/ampel${{ steps.platform.outputs.ext }} https://github.com/carabiner-dev/ampel/rel

───────────────────────────────────────────────────────────────────────────────────── Binary Download Verification ─────────────────────────────────────────────────────────────────────────────────────
  ✗ action.yml : 1 unverified download(s) (no checksum/signature check in file)
    line 2: curl -Lo ${{ inputs.install-dir }}/bin/bnd${{ steps.platform.outputs.ext }} https://github.com/carabiner-dev/bnd/release
```

but it's not obvious to me what that's referring to tbh

potiuk:
> but it's not obvious to me what that's referring to tbh

I think this will be gone when this one https://github.com/apache/infrastructure-actions/pull/800 gets approved and merged - so I need it :) 

The carabiner dev is now recognized as one that actually verifies the downloaad - and #800 is supposed to detect that - but we need to merge it to rebase this one and see if it works.

potiuk:
@pjfanning thanks for the PR. Heads up: I ran `verify-action-build` (the ASF security checker) against `carabiner-dev/actions/ampel/verify@360ffa1…` and it fails with **2 unverified binary downloads**, both inside the nested install actions:

- `carabiner-dev/actions/install/ampel@ba6f58b…` does `curl -Lo …/ampel … && chmod 0755 …` with no sibling checksum / signature / attestation step.
- `carabiner-dev/actions/install/bnd@ba6f58b…` has the same shape for `bnd`.

I manually inspected both action.yml files to confirm — this isn't a tooling false positive. There's no verification of any kind on the downloaded binaries.

I've opened an upstream issue asking carabiner-dev to add verification (gh attestation, `SHA256SUMS`, or cosign): https://github.com/carabiner-dev/actions/issues/51. Could you support it there and add a comment saying you'd like to use this action from ASF infra? An ack from a real downstream consumer would help move it along.

Until that ships we can't approve this action, but it should be a quick fix once they pick a verification mechanism.

pjfanning:
Looks like the maintainer of the carabiner actions is looking at making changes. We can hold off on updating setup-sbt in our Pekko builds for the time being

potiuk:
> Looks like the maintainer of the carabiner actions is looking at making changes. We can hold off on updating setup-sbt in our Pekko builds for the time being

Yep. Looks like everyone is on a spree to fix supply-chain holes :) 

I really think that when we all start pinging and poking each other (esp. our dependencies) - we can nicely help the whole community to get better at security. That's what my Beach Cleaning is all about - and with agents helping us - it's 100% possible.

BTW. This comment abov and comment to the carabiner. Was 100% proposed by my agent - I just pointed it to the PR and guided it a bit what to do.

pjfanning:
@potiuk @raboof v1.2.0 is already out and the CI here passes with it

raboof:
Nice!

I see the "Verify Manunal Allowlist PR" output shows some "hash-pinned, NOT in our approved list" messages for transitive dependencies - I suspect we might need to allowlist those as well.

If that's the case then perhaps those observations should fail the check as well?

raboof (approved):
I'm convinced this action is worth allowlisting, though

potiuk:
> I see the "Verify Manunal Allowlist PR" output shows some "hash-pinned, NOT in our approved list" messages for transitive dependencies - I suspect we might need to allowlist those as well.
> If that's the case then perhaps those observations should fail the check as well?

Of course. PRs to do it are most welcome

PR #803, opened 2026-05-05
[INFRA-27887] Add versions used by flink-docker

snuyanzin:
Apache Flink uses https://github.com/apache/flink-docker
to publish docker images and it uses some docker actions which are not in whitelist
PR to add them here

PR #804, opened 2026-05-05
verify-action-build: extract all action refs from multi-action PRs

potiuk:
## Summary

- `extract_action_refs_from_diff` only seeded the actions.yml key context from `+` lines, so hash additions under pre-existing keys (where the key shows up only as a context line or in the hunk header's tail) were silently dropped — apache/infrastructure-actions#803 has 3 new entries, but pre-fix `--from-pr 803` only returned the one wholly-new key `docker/bake-action`.
- Read the key from `+`/context lines and from `@@ ... @@ <key>:` hunk-header tails. Reset the key on hunk boundaries so stale context can't bleed across unrelated edits. Removed lines (`-`) deliberately don't seed — a deletion block must not leave a context for any later hash.
- 5 new tests: hash-under-existing-key (context line + hunk header), the PR 803 fixture (3 refs), and two negative cases.

## Test plan

- [x] `uv run pytest utils/tests/` — 209/209 pass.
- [x] Real PR 803 diff fed through the extractor returns all 3 refs (was 1 pre-fix):
  - `docker/bake-action@aefd381c…`
  - `docker/build-push-action@ca052bb5…`
  - `docker/setup-qemu-action@c7c53464…`

### DIFF
--- a/utils/tests/verify_action_build/test_pr_extraction.py
+++ b/utils/tests/verify_action_build/test_pr_extraction.py
@@ -122,3 +122,119 @@ class TestExtractActionRefsFromDiff:
         refs = extract_action_refs_from_diff(diff)
         assert len(refs) == 1
         assert "abc123def456789012345678901234567890abcd" in refs[0]
+
+    def test_hash_added_under_existing_key_via_context_line(self):
+        # The most common shape of a real PR diff: the org/repo key is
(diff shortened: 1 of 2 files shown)


dave2wave (approved):
Let's play in production

PR #806, opened 2026-05-06
Add `jbangdev/setup-jbang` action with tag v0.1.1

tiagobento:
## **Why this action is needed for your project**
A new CI / PR checks system introduced in KIE (https://github.com/apache/incubator-kie-drools/commit/eaed2789c5df6acc6366d61b820695aebc364252) needs to run some scripts in order to produce efficient build partitioning for Maven builds. JBang was selected because it doesn't require any additional tooling for developers to install.

## **Any alternatives you've considered**
Installing JBang manually but I would just be duplicating what the action already does. See manual installation instructions [here](https://www.jbang.dev/documentation/jbang/latest/installation.html#universal-all-platforms). And the action's code [here](https://github.com/jbangdev/setup-jbang/blob/v0.1.1/action.yml).

## **Any security concerns you've identified**
None. `jbagdev` is the official organization of the JBang project in GitHub.

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -508,6 +508,9 @@ jarvusinnovations/background-action:
 jasonetco/create-an-issue:
   1b14a70e4d8dc185e5cc76d3bec9eab20257b2c5:
     tag: v2
+jbangdev/setup-jbang:
+  2b1b465a7b75f4222b81426f23a01e013aa7b95c:
+    tag: v0.1.1
 JetBrains/qodana-action:
   89eb4357efd2b52e639f3216e63edaf33b82622b:
     tag: v2025.3.2


potiuk:
@tiagobento thanks for the PR. Heads up: I ran `verify-action-build` (the ASF security checker) against `jbangdev/setup-jbang@2b1b465a…` and it fails with **1 unverified binary download** plus a **pipe-to-shell (curl | sh) — high risk** flag, both pointing at the same line:

```yaml
curl -Ls https://sh.jbang.dev | bash -s - app setup
```

…and the equivalent `iex "& { $(iwr https://ps.jbang.dev) } app setup"` on Windows. The action's repo also has no `LICENSE` file. Manually inspected `action.yml` to confirm — there's no checksum, signature, or attestation step anywhere on the install path.

The good news is that upstream `jbangdev/jbang` already ships SHA256 sums and GPG signatures on every release, so this is a swap, not new infrastructure.

I've opened an issue with a concrete proposal: https://github.com/jbangdev/setup-jbang/issues/16. Could you support it there and add a comment saying you'd like to use this from ASF infra? An ack from a real downstream consumer would help nudge it along.

Until verification lands upstream we can't approve this action — but it should be a quick fix once they pick a mechanism (sha256sum, gpg --verify, or sigstore are all accepted).

tiagobento:
Thanks @potiuk! In the mean time, is it okay if I install a specific version manually and perform the checksums myself, in a temporary custom action? I can tag you in the PR where I do it, so that you can advise on best practices.

potiuk:
> Thanks @potiuk! In the mean time, is it okay if I install a specific version manually and perform the checksums myself, in a temporary custom action? I can tag you in the PR where I do it, so that you can advise on best practices.

Sure. No need to tag me.

potiuk:
Thanks again @tiagobento for the contribution — going to close this out for now. The two blockers from earlier in the thread (`curl -Ls https://sh.jbang.dev | bash -s -` with no checksum/signature/attestation, plus no LICENSE on the `jbangdev/setup-jbang` repo) both need upstream changes before the action can go on the allowlist, and it sounds like the temporary custom-action route in your own project is the right path for now.

Happy to revisit if `sh.jbang.dev` ever ships SLSA attestation or jbangdev publishes a verifiable install path (immutable release + checksums, or `actions/attest-build-provenance`).

PR #808, opened 2026-05-06
verify-action-build: detect orphan tags that ship package.json

potiuk:
## Summary

- `is_source_detached` required `not has_pkg` — the detector only flagged orphan release tags whose tree had **no** `package.json`. But real-world release-tagging workflows commonly **include** `package.json` on the orphan tag (so consumers reading the manifest at runtime, and the action's own `node_modules/`, both resolve) while still excluding the `src/` tree.
- `benchmark-action/github-action-benchmark@v1.22.0` (`a60cea5b…`, surfaced via apache/infrastructure-actions#805) is exactly this shape: tag has `dist/`, `package.json`, `package-lock.json`, `action.yml`, `node_modules/` — and no `src/`. Pre-fix the source-detached fallback never fired, the rebuild ran against a tree with no source, and `canonicalizeUnit.js` showed up as "only in original" in the JS-rebuild diff. After the fix, the fallback resolves the master `release v1.22.0` commit (`c38cce98215c`) and rebuilds from there; all 16 `src/*.js` files are byte-identical.
- Drop the `not has_pkg` requirement. Add a `not has_root_ts_source` guard so root-source actions (`index.ts` next to `action.yml`, no `src/`) aren't false-positived.
- 7 regression tests in `test_release_lookup.py` covering the benchmark-action shape, the original PR #768 shape (no regression), the common `src/`-tree shape, the root-TS-source shape, composite/docker without `dist/`, sub_path, and an empty tree.

## Test plan

- [x] `uv run pytest utils/tests/` — 216/216 pass.
- [x] `prek run --all-files` is clean.
- [x] Re-running `verify-action-build benchmark-action/github-action-benchmark@a60cea5b…`: source-detached fallback fires, rebuilds from `c38cce98215c`, all 16 `src/*.js` files identical.
- [ ] Manual: re-trigger CI on apache/infrastructure-actions#805 once this lands and confirm the previously-failing `JS build verification` check passes.

Note: a single residual finding remains on PR #805 — `scripts/ci_validate_modification.js` is "only in rebuilt" because the publisher's release workflow excludes CI helper scripts from `dist/`, but `tsconfig.build.json`'s `**/*.ts` pattern compiles them. That's a separate concern (lenience for "extra in rebuilt" outputs vs. publisher's curation) — out of scope for this PR.

### DIFF
--- a/utils/tests/verify_action_build/test_release_lookup.py
+++ b/utils/tests/verify_action_build/test_release_lookup.py
@@ -22,6 +22,7 @@ from unittest import mock
 from verify_action_build.release_lookup import (
     format_release_time,
     get_release_or_commit_time,
+    is_source_detached,
 )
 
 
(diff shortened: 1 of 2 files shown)


raboof (approved):
that commit summary could've been a lot clearer I think.

I guess I can see the implementation, but it does seem rather brittle.

potiuk:
> that commit summary could've been a lot clearer I think.

We can definitely write down rules in AGENTS.md to get it less chatty if you do not need that many details. I am happy to add it now. I do not like too much those very long descriptions - in this repo it is not needed, what for example Airflow benefits from to have more details - but luckily "intention is all you need" - it's enough to express it - and we have it.

potiuk:
> that commit summary could've been a lot clearer I think.
> 
> I guess I can see the implementation, but it does seem rather brittle.

https://github.com/apache/infrastructure-actions/pull/815 @raboof -> this should make  PR /commit descriptions shorter.

PR #818, opened 2026-05-07
allowlist: remove browser-actions/setup-geckodriver and /setup-firefox

raboof:
As Tapestry was the only one using them, and they
moved away from it.

https://github.com/apache/infrastructure-actions/pull/813 https://github.com/apache/tapestry-5/pull/58

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -136,14 +136,6 @@ biomejs/setup-biome:
 BobAnkh/auto-generate-changelog:
   '*':
     keep: true
-browser-actions/setup-firefox:
-  5914774dda97099441f02628f8d46411fcfbd208:
-    tag: v1.7.0
-    expires_at: 2026-06-12
-  fcf821c621167805dd63a29662bd7cb5676c81a8:
-    tag: v1.7.1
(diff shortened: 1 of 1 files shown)


dfoulks1:
This LGTM, not 100% sure what the failure is about though.

raboof:
> not 100% sure what the failure is about though.

It couldn't determine what action was added, which makes sense since no action was added :)

I'm OK with that: I'd rather have an incorrectly-red test in this case then an incorrectly-green one when the action is not recognized for some reason.

potiuk (approved):

potiuk:
> I'm OK with that: I'd rather have an incorrectly-red test in this case then an incorrectly-green one when the action is not recognized for some reason.

I can ask my agent to fix it :)

potiuk:
https://github.com/apache/infrastructure-actions/pull/820 -> my aagent did good job :)

potiuk:
Can you rebase @raboof ? Just to check if my merged fix from #820 worked :)

potiuk:
OH... actually I could rebase it - green now. Merging :)

PR #819, opened 2026-05-07
verify-action-build: verify in-tree binaries (attestation / SHA256SUMS)

potiuk:
## Summary

- Add an **In-tree binary check** to `verify-action-build`. Detects actions that ship pre-compiled native binaries directly in the repo (Go cross-compiled like `main-linux-amd64`, plus `*.exe`/`*.dll`/`*.so`/`*.dylib`/`*.jar`/`*.wasm`, etc.) and exec them from a small launcher.
- For each detected binary, verify against upstream provenance via a cascade:
  1. **`gh attestation verify --owner <org>`** — queries the GitHub attestation transparency log (populated by [`actions/attest-build-provenance`](https://github.com/actions/attest-build-provenance)).
  2. **`SHA256SUMS` release asset** — fetches the release's checksum file and SHA256-compares each binary.
  3. **Neither** — hard fail. Action runs opaque code; reject until upstream adds provenance.
- Verified binaries are reported ✓ with the verification mechanism. Unverified binaries fail the action with offending paths named.
- Document the new check in README's security review checklist + add case **F** to the `analyze-action-pr` skill, with `runs-on/action@v2.1.1` (rejected — no provenance) and `v2.1.2` (verified — both attestation and SHA256SUMS) as precedents.

## Test plan

- [x] `uv run pytest utils/tests/` — 236/236 (15 new for in-tree binaries + 6 new for SHA256SUMS parser).
- [x] `prek run --all-files` clean.
- [x] **Reject case** (`runs-on/action@v2.1.1`): no provenance — 3 binaries flagged as unverified, action fails.
- [x] **Verify case** (`runs-on/action@v2.1.2`): upstream landed `actions/attest-build-provenance` + `SHA256SUMS` after our [issue runs-on/action#36](https://github.com/runs-on/action/issues/36) — all 3 binaries verified via `gh attestation verify`, action passes (exit 0).

## Background

Closes the gap surfaced by apache/infrastructure-actions#809 (`runs-on/action@v2.1.1`). The maintainer ([@crohr](https://github.com/crohr)) responded to our upstream ask by adding both verification mechanisms in [runs-on/action#37](https://github.com/runs-on/action/pull/37); v2.1.2 is now fully verifiable, which is what apache/infrastructure-actions#825 (the v2.1.2 dependabot bump) needs from us.


dave2wave:
My approval is contingent on @gmcdonald confirming that `runs-on` has been rejected as a vendor.

If it still needs to be used then I think we will need an allow-list of actions that pass verify. Maybe these are ones we can wildcard.

potiuk:
> My approval is contingent on @gmcdonald confirming that `runs-on` has been rejected as a vendor.
> 
> If it still needs to be used then I think we will need an allow-list of actions that pass verify. Maybe these are ones we can wildcard.

Note that failing the test does not mean rejection. It just means failing the test. It can still be approved.

potiuk:
> My approval is contingent on @gmcdonald confirming that `runs-on` has been rejected as a vendor.
> 
> If it still needs to be used then I think we will need an allow-list of actions that pass verify. Maybe these are ones we can wildcard.

Also - if you look there https://github.com/runs-on/action/issues/36 The maintainer had been very responsive and implemented attestation for building those binaries (cc: @ppkarwasz ) - so we still might **allow** the actions where attestations are present - they do not give us 100% reproducibility - but I think the trust in this case moves from the author to GitHub + review of the build steps. Am I right Piotr?

potiuk:
> My approval is contingent on @gmcdonald confirming that `runs-on` has been rejected as a vendor.
> 
> If it still needs to be used then I think we will need an allow-list of actions that pass verify. Maybe these are ones we can wildcard.

Also ... I added (as a follow-up after the fix implemented by @crohr in https://github.com/runs-on/action/issues/36 -> this check will now only fail if we "can't" verify the binary - wiht either binary attestations or SHASUM approch. This PR has been tested on #825 and it correctly verify the attested binaries - with SLSA attestations of build:

image

Where it fails on previous version of the same bump, where the action did not have attestations:


image


Which is probably exactly what we want.

dave2wave (approved):
Thank you

PR #821, opened 2026-05-07
verify-action-build: widen source-commit window + prefer exact tag match

potiuk:
## Summary

- `resolve_source_commit`'s `published_at + 1min` cutoff missed release-please-style orphan tags where the master `release vX.Y.Z` commit lands seconds **after** the tag is published. benchmark-action/github-action-benchmark v1.22.1 (apache/infrastructure-actions#805) hits this: tag published at `10:36:23Z`, master release commit at `10:37:24Z` — 1 second past the old cutoff. Resolver fell back to v1.22.0's release commit; rebuild produced v1.22.0's dist; verification reported "DIFFERENCES DETECTED" against v1.22.1's published dist.
- Widen the search window to `±2h` around the anchor and add a strong-preference match on the **exact tag name** in the commit message (`release v1.22.1` matches tag `v1.22.1`; also matches `release 1.22.1` without the `v` prefix). Falls back to the existing generic release-marker on miss.
- 6 new tests in `test_release_lookup.py` covering: orphan-tag-pushed-before-master-release; multiple "release X" commits in window (exact match wins); release-please's no-`v`-prefix message variant; release-marker fallback; empty-commit-list and missing-release edge cases.

## Test plan

- [x] `uv run pytest utils/tests/` — 229/229.
- [x] `prek run --all-files` clean.
- [x] Smoke: `verify-action-build benchmark-action/github-action-benchmark@52576c92…` (PR #805's SHA) — was failing on origin/main; with the fix the source-detached fallback resolves to `67cf262e8c79` ("release v1.22.1") and **all 16 `src/*.js` files are byte-identical**. Exit 0.

Generated-by [Claude Code](https://claude.com/claude-code).

PR #822, opened 2026-05-07
verify-action-build: ignore IDE config dirs in vendored node_modules diff

potiuk:
## Summary

- Some upstream npm packages ship author IDE config trees (case-in-point: `node_modules/tunnel/.idea/*` in benchmark-action's vendored deps surfaced via apache/infrastructure-actions#805). These leak into a fresh `npm ci` but not the scrubbed vendored copy, producing noisy "Files only in rebuilt" output that's not security-relevant.
- Extend `diff_node_modules`'s existing `noisy_dirs` filter (already covers `.cache`, npm metadata) with `.idea`, `.vscode`, `.vs`, `.fleet`, `.zed`.
- 2 new tests in `test_diff_node_modules.py`: IDE dirs ignored on both sides; non-IDE dot-dirs (`.bin`) with real content still compared.

## Test plan

- [x] `uv run pytest utils/tests/` — 225/225.
- [x] `prek run --all-files` clean.

Note: PR #805's original symptom turned out to be entirely caused by the wrong source-commit resolution (fixed in #821). With #821, `tunnel/.idea/*` doesn't appear at all because the rebuild uses v1.22.1's lock file, which pins a `tunnel` version that doesn't ship `.idea` files in its tarball. This PR is **defensive** — same drift can recur on other actions/versions, and filtering IDE config dirs from the comparison is harmless.

### DIFF
--- a/utils/tests/verify_action_build/test_diff_node_modules.py
+++ b/utils/tests/verify_action_build/test_diff_node_modules.py
@@ -100,6 +100,69 @@ class TestDiffNodeModules:
         )
         assert result is True
 
+    def test_ide_config_dirs_ignored(self, tmp_path):
+        # Some upstream npm packages ship author IDE config trees
+        # (``node_modules/tunnel/.idea/*`` was the case-in-point in
+        # benchmark-action's vendored deps).  These are not runtime code
+        # and shouldn't break verification when they leak into one side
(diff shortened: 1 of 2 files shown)


dave2wave (approved):
Pragmatic response

PR #833, opened 2026-05-12
Add azure/login action with version v3.0.0

tisonkun:
Used by https://github.com/apache/opendal-reqsign/actions/runs/25707523312

And this should be the official azure actions that can be trusted.


potiuk:
All check look good for it!

potiuk (approved):

PR #846, opened 2026-05-15
Allow matlab-actions/run-tests v3.1.1

kevingurney:
To address: [#49962](https://github.com/apache/arrow/pull/49962), we would like to allow `v3.1.1` of the `matlab-actions/run-tests` action.

**Note**: `v3.1.1` is *not* a  Prerelease, so we don't expect to run into the "commit churn" which was addressed by #735 for this version. On behalf of the MathWorks CI Team - our sincere apologies for the  inconvenience this issue caused.

**Note**: It is unclear what is preventing `matlab-actions/run-tests` from eventually ending up in the same state again if a future upgrade is also a Prerelease. Does some kind of additional logic need to be added to the infrastructure to "ignore" prereleases for Dependabot upgrades?

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -687,6 +687,8 @@ matlab-actions/run-tests:
   4be3345d41da8b1bf8cc5cb01a5e19c4611cb15d:
     tag: v3.0.0
     expires_at: 2026-07-05
+  4c375221b861dd3f7ea9b31704a3cf45d35073f1:
+    tag: v3.1.2
 matlab-actions/setup-matlab:
   aa8bbc7b76daa63c5d456d1430cbd6cb5b626ab4:
     tag: v2.7.0


potiuk:
Hi @kevingurney — thanks for the bump.

Running our automated allowlist check (`verify-action-build`) on v3.1.1 turns up one issue we'd want to resolve before approving:

> ✗ `dist/bin/win64/run-matlab-command.exe`: no SLSA attestation and release v3.1.1 has no SHA256SUMS

The JS rebuild matches cleanly (✓), no curl-downloads, lock file present. The blocker is just the four `dist/bin/*/run-matlab-command*` launchers that v3.1.x now commits to the tree — they're opaque to any rebuild check, and `matlab-actions/run-tests` releases don't ship `SHA256SUMS` or SLSA attestations, so there's no way for us to tie the bytes back to the source/workflow that produced them.

(Side note: vendoring the binaries into `dist/` is actually a security improvement over fetching them via `common-utils` at install time — the SHA-pin now freezes the bytes. We just need the upstream provenance hook to close the verification loop.)

I've opened matlab-actions/run-tests#94 proposing `actions/attest-build-provenance` (or `SHA256SUMS` as a release asset) as the fix — same one-line pattern that [runs-on/action#37](https://github.com/runs-on/action/pull/37) used. Since you're at MathWorks, you may have a faster path to that team than a drive-by issue does — if you could support the upstream request as a downstream consumer, that'd unblock this PR (and every future `matlab-actions/*` bump).

Holding approval until upstream provenance lands.

kevingurney:
@potiuk - thanks so much for the detailed response and for opening https://github.com/matlab-actions/run-tests/issues/94!

I will get in touch with the MATLAB Actions Team at MathWorks to see if we can push on addressing this limitation.

I really appreciate you taking the time to investigate this.

potiuk:
> I really appreciate you taking the time to investigate this.

You are most welcome !

potiuk:
Good news @kevingurney — MathWorks moved fast on your push. @davidbuzinski shipped v3.1.2 a few days after we filed `matlab-actions/run-tests#94`, with full SLSA build provenance on the in-tree `dist/bin/*/run-matlab-command*` launchers via `actions/attest-build-provenance`. Just re-ran `verify-action-build` locally against `4c375221b861` (v3.1.2) — clean across the board, including the in-tree binary check that was the original blocker.

Going to update this PR to point at v3.1.2 instead of v3.1.1 (v3.1.1 still ships the unattested binaries; bumping straight to v3.1.2 is the better landing spot and keeps the allow list at the latest stable). Thanks for your patience and for prodding MathWorks — this unblocked the policy permanently.

potiuk (approved):
Verified locally with `uv run utils/verify-action-build.py --ci matlab-actions/run-tests@4c375221b861` — clean across the board (in-tree binary check ✓ via the new SLSA attestations from MathWorks), and the PR's own `verify` job in CI just confirmed the same against the rebased branch. Approving + merging.

kevingurney:
Thanks so much for the help, @potiuk! Glad we were able to get this resolved quickly!

PR #850, opened 2026-05-17
verify-action-build: detect MATLAB platform-named binary dirs

potiuk:
## Summary

- `_looks_like_in_tree_binary` flagged `dist/bin/win64/run-matlab-command.exe` (via its `.exe`) but missed the three Unix siblings at `dist/bin/{glnxa64,maca64,maci64}/run-matlab-command` — they have no extension and don't fit the cross-compile `<name>-<os>-<arch>` regex.
- Add a parent-directory rule: if the immediate parent dir is a known MATLAB platform identifier (`glnxa64`, `maca64`, `maci64`), the file is treated as a launcher binary. Windows already worked via extension.
- Pin regression tests against the actual `matlab-actions/run-tests@v3.1.1` shape so future bumps can't regress the catch.

Surfaced while triaging apache/infrastructure-actions#846 (upstream provenance ask: matlab-actions/run-tests#94).

## Test plan

- [x] `uv run pytest utils/tests/` — 273 passed
- [x] `prek run --all-files` — clean

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -1302,6 +1302,32 @@ class TestLooksLikeInTreeBinary:
         assert _looks_like_in_tree_binary("dist/licenses.txt") is False
         assert _looks_like_in_tree_binary("licenses.txt") is False
 
+    def test_matlab_platform_dir_naming(self):
+        # MATLAB's launcher convention: dist/bin/<platform>/run-matlab-command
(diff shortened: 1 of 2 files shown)


ppkarwasz (approved):
LGTM, although having files in a folder called `bin`, seems like a pretty good hint that these are binary too.

potiuk:
> LGTM, although having files in a folder called `bin`, seems like a pretty good hint that these are binary too.

Unless they are scripts :)

PR #851, opened 2026-05-17
verify-action-build: treat `Authorization: token/Bearer` headers as data-fetch signal

potiuk:
## Summary

- `manusa/actions-setup-minikube@v2.17.0` has a thin `axios` wrapper in `src/github.js` that fetches GitHub release metadata. The file is a transport-layer helper (no data-parse markers, no binary-handling markers) so `_file_is_pure_data_fetch` didn't exempt it and the `axios()` call landed in the unverified-download list.
- An `Authorization` header with `token ` or `Bearer ` is an unambiguous "authenticated API call" signal — public release-asset downloads from S3 / CDN / GitHub Releases don't use bearer auth. Add it to `_JS_DATA_PARSE_PATTERNS` so transport-layer API helpers stop firing.
- Real binary handlers in the same file still trip `_JS_BINARY_HANDLE_PATTERNS` and override the exemption — test asserts this.

Surfaced while triaging apache/infrastructure-actions#848.

## Test plan

- [x] Regression fixture using the verbatim `src/github.js` from `manusa/actions-setup-minikube@v2.17.0`.
- [x] Negative test: same file plus `tc.downloadTool` + `tc.extractTar` → both downloads still flagged.
- [x] `uv run pytest utils/tests/` — 273 passed.
- [x] `prek run --all-files` — clean.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -1085,6 +1085,73 @@ const data = JSON.parse(await response.readBody())
 """
         assert _file_is_pure_data_fetch(content) is True
 
+    # The exact pattern that triggered the false positive on PR #848
+    # (manusa/actions-setup-minikube@v2.17.0): a thin axios wrapper that
+    # sets ``headers.Authorization = `token ${githubToken}`` and returns
+    # the raw response.  The file is a transport-layer helper — no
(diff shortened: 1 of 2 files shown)


ppkarwasz (approved):
LGTM

Issue #852, opened 2026-05-18
Investigate tag-based pinning for sibling-calling actions (immutable releases)


potiuk:
## Context

PR #831 surfaced a scaling concern raised by @ppkarwasz: actions that call sibling sub-actions (e.g. `carabiner-dev/actions/ampel/verify` calling `carabiner-dev/actions/install/*`) currently force us to allowlist **two SHAs per release** — the current version plus the previous version that the sibling refs resolve to (per github/community discussion #26245). This scales as `2 × siblings × versions` and is wasteful against the 1000-pattern org cap.

ppkarwasz's full argument: https://github.com/apache/infrastructure-actions/pull/831#issuecomment-4474700884

## Proposal under evaluation

1. **Ask upstream authors (starting with @puerco / `carabiner-dev`) to enable immutable releases** so their version tags can no longer be moved post-publication.
2. **Ask the same authors to pin sibling sub-action refs by tag** instead of by previous-version SHA.
3. **Teach our verifier (`gateway/gateway.py` and friends) to accept tag-pinning when the tag is provably immutable**, ideally with a pattern like `owner/repo/sub-action@vX.Y.Z` rather than `@<sha>`.

## Open questions to resolve before implementing (3)

- **Path-segment wildcards.** Does GitHub's org allowlist actually expand patterns like `carabiner-dev/actions/*@v1.2.0`? All current wildcards in `approved_patterns.yml` are version-side (`@*`). If path wildcards aren't supported, the saving shrinks to "one pattern per (sub-action, version)".
- **Immutability re-verification.** GitHub's per-release "immutable" flag can be disabled by the repo owner on future releases. Either:
  - Re-check immutability at every audit run (network cost, rate-limit cost, requires API endpoint), or
  - Snapshot the flag at approval time and accept silent erosion of the guarantee.
- **Policy update.** README currently says "Always pin actions to exact commit SHAs, never use tags or branch references." Tag-pinning under immutability is a controlled exception — needs to be documented as such, not as a blanket relaxation.
- **Threat model delta.** What's the additional blast radius if a repo turns off immutable releases after approval and an attacker then moves a tag we've trusted? Compare against current SHA-pinning baseline.

## Non-goals

- Replacing the current SHA-pinning policy as the default. SHA-pinning remains the strictly safer primitive; this is about a documented escape hatch for the sibling-action case.

## Suggested next step

[...]

PR #853, opened 2026-05-18
allowlist: add carabiner-dev install/{download-and-verify,ampel-bootstrap} (levels 2-3)

potiuk:
## Summary

Follow-up to #831 — that PR added the level-1 transitive siblings `install/{ampel,bnd}@v1.1.7`, but the sibling-call chain rooted at `carabiner-dev/actions/ampel/verify@v1.2.0` is actually four levels deep:

```
ampel/verify @ v1.2.0 (e0e3b81…)
  └── install/{ampel,bnd} @ v1.1.7 (2a11d59…)             [added by #831]
        └── install/download-and-verify (6022a06…)        [this PR]
              └── install/ampel-bootstrap (0a075bb…)      [this PR]
```

`ampel-bootstrap` is a leaf — no further sibling refs.

The hourly `check-for-transitive-failures` workflow has continued to fail post-#831 with `carabiner-dev/actions/install/download-and-verify@6022a065d6420de5d86333ecfb2b25c57f84b699 is not allowed` (e.g. [run 26062684716](https://github.com/apache/infrastructure-actions/actions/runs/26062684716)). After this PR the next failure (if any) would be on `ampel-bootstrap`, which is also pre-emptively added here.

## Test plan
- [ ] On merge, `update_composite_action.yml` regenerates `approved_patterns.yml` + the dependabot composite from `actions.yml`.
- [ ] ASF Infra's allowlist sync picks up the two new SHAs; the next hourly `check-for-transitive-failures` run is green.

## Related
- #831 — original (level-1) fix
- #852 — tracking issue for the underlying sibling-chain pattern-explosion (pp's scaling concern, also see [his comment](https://github.com/apache/infrastructure-actions/pull/831#issuecomment-4474700884))
- carabiner-dev/actions#57 — upstream ask to enable immutable releases + tag-pin siblings

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -181,18 +181,16 @@ carabiner-dev/actions/install/ampel:
     # transitive dep of carabiner-dev/actions/ampel/verify @ v1.2.0
     tag: v1.1.7
 carabiner-dev/actions/install/ampel-bootstrap:
+  # level-3 transitive: called by install/download-and-verify; commit is untagged
+  # (per upstream: "bootstrap rotated to ampel v1.2.1")
   0a075bb75a68646d05f99c85cbbf2be40dd8e442:
-    # level-3 transitive: called by install/download-and-verify
-    # (per upstream: "bootstrap rotated to ampel v1.2.1")
-    tag: v1.1.7
 carabiner-dev/actions/install/bnd:
   2a11d59a135c5e291f305f249a92ad7903e3ee0f:
     # transitive dep of carabiner-dev/actions/ampel/verify @ v1.2.0
     tag: v1.1.7
 carabiner-dev/actions/install/download-and-verify:
+  # level-2 transitive: called by install/{ampel,bnd} @ v1.1.7; commit is untagged
   6022a065d6420de5d86333ecfb2b25c57f84b699:
-    # level-2 transitive: called by install/{ampel,bnd} @ v1.1.7
-    tag: v1.1.7
 carloscastrojumo/github-cherry-pick-action:
   503773289f4a459069c832dc628826685b75b4b3:
     tag: v1.0.10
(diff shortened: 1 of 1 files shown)


actions.yml
```diff
@@ -184,6 +184,15 @@ carabiner-dev/actions/install/bnd:
   2a11d59a135c5e291f305f249a92ad7903e3ee0f:
     # transitive dep of carabiner-dev/actions/ampel/verify @ v1.2.0
     tag: v1.1.7
+carabiner-dev/actions/install/download-and-verify:
+  6022a065d6420de5d86333ecfb2b25c57f84b699:
+    # level-2 transitive: called by install/{ampel,bnd} @ v1.1.7
+    tag: v1.1.7
```
ppkarwasz: This is not tag `v1.1.7`. Apparently the innermost commits are untagged.

actions.yml
```diff
@@ -184,6 +184,15 @@ carabiner-dev/actions/install/bnd:
...
+carabiner-dev/actions/install/ampel-bootstrap:
+  0a075bb75a68646d05f99c85cbbf2be40dd8e442:
+    # level-3 transitive: called by install/download-and-verify
+    # (per upstream: "bootstrap rotated to ampel v1.2.1")
+    tag: v1.1.7
```
ppkarwasz: Same here: let's leave the comment, but remove the tag.

ppkarwasz (changes requested):
I had a feeling, there were more than one level of indirection…

actions.yml
potiuk: Yeah. Tag seems to come from outer action..

potiuk: Good catch — fixed in 715572c. Dropped `tag: v1.1.7` on both untagged entries, moved the call-chain note up to the action-key line so it survives the sort-actions-yml hook (standalone comments under an empty SHA sub-map get stripped on sort).

actions.yml
potiuk: Done in 715572c — same shape as the download-and-verify entry.

ppkarwasz (approved):

PR #876, opened 2026-05-22
workflows: fix allowlist race, inline cap check, drop PAT (#866)

### LINKED ISSUE #866: race condition in allowlist / action updates
what _should_ have happened:

* dependabot commit 1cfd7331df7da6a38a05eafeb679cc0faaba8667 updated `.github/actions/for-dependabot-triggered-reviews/action.yml`
* `.github/workflows/update_actions.yml` updates `actions.yml` and `approved_patterns.yml`

but what _actually_ happened:

* dependabot commit 1cfd7331df7da6a38a05eafeb679cc0faaba8667 updated `.github/actions/for-dependabot-triggered-reviews/action.yml`
* manual commit ba7238de2e4e8c7c89884e1089a83372b47e8945 updated `actions.yml`
* `.github/workflows/update_composite_action.yml` commit d11241348d947c569a8f16dd8c503bbc7df3ad43 updated `approved_patterns.yml` and overwrote(!) `.github/actions/for-dependabot-triggered-reviews/action.yml`

so a race condition. we should think about how we can avoid this - possibly we can combine the two actions into one, making sure we always first add information from `.github/actions/for-dependabot-triggered-reviews/action.yml` to `actions.yaml` and only after that do the update in the opposite direction.

_Originally posted by @raboof in https://github.com/apache/infrastructure-actions/issues/857#issuecomment-4509516191_

potiuk:
## Summary

Consolidates `update_actions.yml` (composite → actions.yml) and
`update_composite_action.yml` (actions.yml → composite) into a single
`update_allowlist.yml`. The previous pair ran in separate concurrency
groups and could overlap, silently clobbering one direction's edit —
see #866 for raboof's reconstruction of how a dependabot bump got
downgraded.

The combined workflow:

- Triggers on push to `main` when either file changes
- Single concurrency group so only one run touches the allowlist at a time
- Always runs both directions: merge composite → actions.yml (additive),
  then regenerate composite + approved_patterns.yml from actions.yml
- Checks out latest `main` HEAD (not the trigger SHA) so a queued run
  sees the prior run's commit
- Rebases and retries on push failure (the sync script is idempotent)

Also folds the former `check_approved_limit.yml` in as an inline step
(`Check approved actions count`) that runs after regeneration and
before commit/push, so an over-cap `approved_patterns.yml` never lands
on `main`. With no downstream push-triggered workflow left to fire,
the workflow now authenticates with the default `GITHUB_TOKEN` — the
PAT (`ALLOWLIST_WORKFLOW_TOKEN`) is gone, and so is the "Print token
details" diagnostic.

README's four mermaid diagrams + prose are rewritten to label edges
with **job names** (`update`, `verify`, `clean_up`) rather than
workflow filenames; the filenames stay in the prose so the docs are
still grep-friendly.

Fixes #866

## Test plan

- [x] `uv run pytest gateway/` — 8 passed
- [x] `uv run pytest` (full suite) — passing (the pre-existing pytest
  collection failure on `utils/check_cache_settings/test_audit_actions_cache.py`
  is fixed by the `c81c85e` hotfix on main, picked up via rebase)
- [x] Local sanity: `grep -c '^- ' approved_patterns.yml` = 342 (well
  under the 800 threshold)
- [ ] After merge: trigger via `workflow_dispatch` on an unchanged
  tree to confirm no-op behavior (sync produces no diff → early-exit,
  cap step still runs and reports the count)
- [ ] Watch the next dependabot bump merge to confirm `actions.yml` +
  composite + `approved_patterns.yml` all update from the single
  workflow with `GITHUB_TOKEN`


.github/workflows/update_allowlist.yml
```diff
@@ -0,0 +1,167 @@
...
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: true
+          token: ${{ secrets.ALLOWLIST_WORKFLOW_TOKEN || github.token }} # zizmor: ignore[secrets-outside-env]
```
ppkarwasz: What is the purpose of the PAT here? Recursive behavior?

This workflow modifies the same file (`actions.yml`) that was used to trigger it. Are we sure the modifications are eventually idempotent?

We should probably add a comment here.

.github/workflows/update_allowlist.yml
```diff
@@ -0,0 +1,167 @@
...
+          g.update_workflow(composite, actions)
+
+          # 3. Regenerate the approved patterns from actions.yml.
+          g.update_patterns(patterns, actions)
+          PYEOF
+
+      - name: Commit and push changes
+        if: ${{ github.event_name != 'pull_request' }}
+        env:
+          GH_TOKEN: ${{ secrets.ALLOWLIST_WORKFLOW_TOKEN || github.token }} # zizmor: ignore[secrets-outside-env]
```
ppkarwasz: What is the purpose of this PAT?

`git` has already been configured with the recursive token above.

.github/workflows/update_allowlist.yml
potiuk: Good catch — fixed in d63ca67 with both an inline comment at this spot (and the other use site below) and a `NOTE` callout in the README under "Updating version of already approved action".

Short answer: the PAT is there so that when this workflow's commit lands on `main`, the push re-triggers downstream push-triggered workflows like `check_approved_limit.yml` (which listens on `approved_patterns.yml`). A push authenticated with the default `GITHUB_TOKEN` is deliberately blocked from creating a new workflow run — per the [GitHub Actions docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow), *"when you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN` will not create a new workflow run"*. A PAT (or GitHub App installation token) bypasses that, so the chain fires.

On idempotency: yes, the sync script is idempotent and the workflow's own `paths` filter on `actions.yml` + the composite means a self-push could in principle re-trigger this very workflow. The `git diff --quiet ... && exit 0` early-exit on line 126 catches the no-op case, so a second run on the rebased tree just terminates without committing. The rebase-and-retry loop (lines 139–165) handles the genuine race where another workflow (e.g. `remove_expired.yml`) advanced `main` while we were computing.

.github/workflows/update_allowlist.yml
potiuk: Same reason as [the comment on line 61](https://github.com/apache/infrastructure-actions/pull/876/files#r3292528699) — the push at the end of this step needs to re-trigger downstream push-triggered workflows, which a `GITHUB_TOKEN`-authored push doesn't do. Both use sites now have an inline comment pointing at the docs (d63ca67).

potiuk:
Update on the PAT discussion above: rather than document the PAT, I dug into what actually downstream-triggered on the post-merge push and found it was exactly one workflow — `check_approved_limit.yml` — guarding the 800/1000 cap on `approved_patterns.yml`. Folding that check into `update_allowlist.yml` as an inline step (running after regeneration and before commit/push, so an over-cap state never lands on main) removes the last reason for the PAT. So this PR now:

- inlines the cap check into the `update` job
- deletes `check_approved_limit.yml`
- drops both `ALLOWLIST_WORKFLOW_TOKEN` uses + the `Print token details` diagnostic step
- authenticates with the default `GITHUB_TOKEN`

Also took the chance to relabel the four mermaid diagrams + prose in the README to use **job names** (`update`, `verify`, `clean_up`) instead of workflow filenames, per your earlier framing. Filenames stay in the prose for grep-ability.

Force-pushed the change as one commit on top of the rebase. Idempotency / self-trigger concerns from your line-61 comment are unchanged — the early-exit at line 137 still catches the no-op case if the workflow's own push re-triggers it (which under `GITHUB_TOKEN` it won't anyway).

potiuk:
I did not like PAT either, but yeah by having a single workflow seems we can get rid of it :)

potiuk:
This ont should be good to go :)

.github/workflows/update_allowlist.yml
```diff
@@ -0,0 +1,172 @@
...
+
+          import gateway as g
+
+          composite = ".github/actions/for-dependabot-triggered-reviews/action.yml"
+          actions = "actions.yml"
+          patterns = "approved_patterns.yml"
+
+          # Always run both directions. The previous two-workflow split
+          # raced on overlapping pushes; running both here means the
+          # outcome is the same regardless of which file triggered us.
```
raboof: I don't find it helpful to describe what the workflow 'no longer does', if needed that can be found in the history

.github/workflows/update.yml
```diff
@@ -0,0 +1,172 @@
...
+# don't drop a dependabot bump or a manual actions.yml edit.
+concurrency:
+  group: "${{ github.ref }}-update-allowlist"
+  cancel-in-progress: false
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
raboof, line 65: this feels a bit icky since this workflow is also triggered on `pull_request`, but GitHub correctly downgrades the permissions in that case, so it's OK.

README.md
```diff
@@ -119,41 +119,39 @@ graph LR
 
     human-->actions
     dependabot-->composite
-    dependabot-.verified by.-verify["verify_dependabot_action.yml<br/>(rebuild &amp; diff)"]
+    dependabot-.verified by.-verify["<b>verify</b> job<br/>(rebuild &amp; diff)"]
 
-    composite=="update_actions.yml<br/>(on merge)"==>actions
-    cron=="remove_expired.yml"==>actions
+    composite=="<b>update</b> job<br/>(on merge)"==>actions
+    cron=="<b>clean_up</b> job"==>actions
```
raboof: I find this adds more confusion than it helps: 'remove expired' was clearer than 'clean up' and it adds unnecessary indirection.

If anything we should rename the `clean_up` job in the `remove_expired` workflow to also be called `remove_expired`.

README.md
```diff
@@ -119,41 +119,39 @@ graph LR
...
+Solid arrows (`==>`) are regeneration edges — the "source → generated" flows that keep `actions.yml`, `approved_patterns.yml` and the dependabot composite in sync. Thin arrows feed the pipeline with new content (human or Dependabot PRs, cron), and dotted arrows are observer jobs that verify rather than mutate. Bold labels are job names (rather than workflow filenames) — `update` lives in `update_allowlist.yml`, `verify` in `verify_dependabot_action.yml` / `verify_manual_action.yml`, `clean_up` in `remove_expired.yml`.
 
 > [!NOTE]
-> `check_approved_limit.yml` guards the whole pipeline: the org-wide allow list has a hard cap of 1000 entries, and this workflow fails at 800 to give maintainers room to clean up before hitting the wall.
+> The 800/1000-entry cap on `approved_patterns.yml` is enforced as a step inside the `update` job (formerly the separate `check_approved_limit.yml`). It runs after regeneration and before commit/push, so an over-cap state never lands on `main`. Folding the check in here also lets the workflow push with the default `GITHUB_TOKEN` — no PAT, no downstream-trigger requirement.
```
raboof: again I don't think it's helpful to describe the 'former' behaviour

.github/workflows/update_allowlist.yml
```diff
@@ -0,0 +1,172 @@
...
+# overlapping pushes, silently overwriting one direction's edit. This
+# single workflow always runs both directions in order, so neither edit
+# is lost regardless of which file triggered the run. See #866.
+#
+# Also subsumes the former `check_approved_limit.yml`: the 800/1000-entry
+# cap on approved_patterns.yml is now enforced inline (see the "Check
+# approved actions count" step). Folding it in here means no downstream
+# push-triggered workflow needs to fire after this workflow commits, so
+# the workflow can authenticate with the default GITHUB_TOKEN — no PAT.
+name: Update Allowlist
```
raboof: this updates the allowlist _and_ the dependabot action. I think just `update.yml` would actually be clearer.

raboof (approved):
ok, but some comments for clarity

.github/workflows/update_allowlist.yml
potiuk: Done in 9bcf6ee — stripped the "Replaces the former pair..." and "Also subsumes the former `check_approved_limit.yml`..." paragraphs from the header, and reworded the inline Python comment + cap-check step comment so they describe only current behavior.

.github/workflows/update.yml
potiuk: Added an inline comment on the `permissions:` block in 9bcf6ee that spells out what you noted — `contents: write` is needed for the push step on push/dispatch, and on `pull_request` GitHub downgrades the `GITHUB_TOKEN` to read-only regardless of what's declared here, so PRs from forks can't acquire write access via this workflow.

README.md
potiuk: Went with the second half of your suggestion in 9bcf6ee: renamed the `clean_up` job in `remove_expired.yml` to `remove_expired`, then updated both mermaid diagrams and all three prose mentions in the README so they use the job name `remove_expired` consistently.

README.md
potiuk: Stripped the "formerly the separate `check_approved_limit.yml`" parenthetical from the NOTE in 9bcf6ee, and reworded the trailing sentence so it explains the current shape (no PAT, no downstream-trigger requirement) without leaning on what used to be there.

.github/workflows/update_allowlist.yml
potiuk: Renamed via `git mv` in 9bcf6ee: `.github/workflows/update_allowlist.yml` → `.github/workflows/update.yml`. Display name bumped to `Update Allowlist and Composite Action` for the same reason. All README references, the workflow's own `paths:` filter, and the `Generated by` commit-message trailers updated to match.

PR #892, opened 2026-05-27
hotfix: restore ALLOWLIST_WORKFLOW_TOKEN PAT in update.yml (branch protection)

potiuk:
## Summary

Restores the `ALLOWLIST_WORKFLOW_TOKEN` PAT on the `update.yml` workflow's checkout `token:` and commit-step `GH_TOKEN` env. Removing it in #876 broke every subsequent run on `main` because `main` is a protected branch and the default `GITHUB_TOKEN` can't push to it.

## What's broken on main right now

Three `update.yml` runs have failed since #889 merged at 22:37 UTC, each with:

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
```

The runs:
- [26542876998](https://github.com/apache/infrastructure-actions/actions/runs/26542876998) — after #889 (configure-aws-credentials v6.1.2)
- [26543850336](https://github.com/apache/infrastructure-actions/actions/runs/26543850336) — after #846 (matlab-actions/run-tests v3.1.2)
- [26543947548](https://github.com/apache/infrastructure-actions/actions/runs/26543947548) — after #886 (zizmor suppression)

Because the push failed, `actions.yml` and `approved_patterns.yml` were never updated with v6.1.2 of `configure-aws-credentials` or v3.1.2 of `run-tests`. The composite is at the new versions, but the source-of-truth file and the org allow list are stuck behind.

## Why #876's PAT removal was wrong

#876 argued that the PAT was only needed to re-trigger the downstream `check_approved_limit.yml` workflow, and that inlining the cap check into `update.yml` removed that need so `GITHUB_TOKEN` could take over. That reasoning missed a second, more important role of the PAT: **branch-protection bypass for the automated commit on `main`**. The default `GITHUB_TOKEN` is blocked by branch protection rules (`GH006`); the PAT was configured with bypass rights for exactly this workflow's commit.

## Changes

- **`.github/workflows/update.yml`** — restore `${{ secrets.ALLOWLIST_WORKFLOW_TOKEN || github.token }}` fallback on both the checkout `token:` input and the commit step's `GH_TOKEN` env. Add inline comments at both use sites explaining the protected-branch reason (rather than the obsolete downstream-trigger rationale). Update the file header note to match.
- **`README.md`** — update the NOTE callout under "Pipeline Overview" so the docs reflect the real reason the PAT is needed.

## Self-healing on merge

When this PR merges, `update.yml` re-runs on the resulting push, the new PAT-authenticated commit step succeeds, and the workflow back-fills `actions.yml` + `approved_patterns.yml` with the v6.1.2 / v3.1.2 entries automatically. No manual cleanup needed.

## Test plan

- [x] `git diff` review — only `update.yml` token paths + comments + README NOTE changed.
- [ ] After merge: confirm the `update.yml` post-merge run pushes successfully (no `GH006`) and that `actions.yml` + `approved_patterns.yml` pick up the v6.1.2 + v3.1.2 entries on the next commit.
- [ ] Watch the next dependabot bump merge to confirm the full pipeline works end-to-end with the restored PAT.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -26,7 +26,9 @@
 # Also enforces the 800/1000-entry cap on `approved_patterns.yml` inline
 # (see the "Check approved actions count" step) — running after
 # regeneration and before commit/push, so an over-cap state never lands
-# on `main`. Pushes use the default `GITHUB_TOKEN`; no PAT.
+# on `main`. The commit-and-push step uses the `ALLOWLIST_WORKFLOW_TOKEN`
+# PAT because `main` is a protected branch; see the comment on the
+# `token:` input below for the full rationale.
 name: Update Allowlist and Composite Action
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


potiuk:
OK. So it seems we still need to use PAT ... @ppkarwasz

potiuk:
Quick ping @raboof @ppkarwasz @dfoulks1 — hotfix for a regression I introduced in #876. Every `update.yml` run on `main` has been failing with `GH006: Protected branch update failed` since #889 merged ~30 min ago, so the workflow's back-fill of `actions.yml`/`approved_patterns.yml` with the recent v6.1.2 + v3.1.2 bumps never landed. Diff is small (workflow token paths + inline comments + a README sentence), CI green. Would appreciate a quick approval so the workflow can re-trigger and self-heal.

potiuk:
Gentle nudge @raboof @ppkarwasz @dfoulks1 — this is still blocking `update.yml` on `main` (every scheduled run since #889 merged is failing with `GH006`). CI's green, diff is small. Would really appreciate a quick look so the workflow can self-heal on the next tick.

raboof (approved):
ah of course

potiuk:
> ah of course

Thank you :)

PR #907, opened 2026-06-02
Allow newer rubygems/release-gem

erickguan:
Adding a new version. Since the upstream doesn't have a new tag, just use `keep`.


potiuk:
Safe to merge: allowlist-only addition of `rubygems/release-gem@f0d7faff` with `keep: true`, pending an upstream release tag (rubygems/release-gem#36) — matches how other not-yet-tagged actions are handled here. All checks green, including `verify`.

potiuk (approved):
Reviewed: green CI, low-risk change as described in the merge note. Approving to merge.

PR #912, opened 2026-06-05
verify: accept validate* checksum helpers as JS download verification (unblocks #910)

potiuk:
## Summary
`verify-action-build`'s binary-download check false-flagged the routine `astral-sh/setup-uv` bump in #910 as an unverified download.

The flagged line is `src/download/download-version.ts:128`:

```ts
const downloadPath = await tc.downloadTool(downloadUrl, undefined, githubToken);
await validateChecksum(checksum, downloadPath, arch, platform, version);
```

The download **is** verified — `validateChecksum` (imported from the sibling `./checksum/checksum`) SHA-256s the artifact against a user-provided checksum or the built-in `KNOWN_CHECKSUMS` table. v8.2.0 just extracted that validation into the sibling module, moving the `createHash` token out of `download-version.ts` and tripping the same-file heuristic (which only knew `verify*`/`compute*`/`calculate*` helper names).

## Fix
Add the `validate*` checksum-helper family to `_JS_VERIFICATION_PATTERNS`. The call name is the in-file evidence that survives the refactor.

## Test plan
- [x] Regression tests in `test_security.py` (bare names + a faithful trim of the real setup-uv download→validate sequence).
- [x] `uv run pytest utils/tests/` → 278 passed.
- [x] `verify_action_build --from-pr 910` → exit 0, "verification present in file".
- [x] prek clean.

Case **E** in the `analyze-action-pr` skill — JS analogue of #800 (sibling `sha256sum -c` counts as verify).

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -1283,6 +1283,42 @@ const createHash = customLibrary.createHash('blake2')
             content = f"const sha = {func_name}(downloadPath)\n"
             assert self._has_verification(content) is True, func_name
 
+    def test_validate_checksum_function_name_recognized(self):
+        # astral-sh/setup-uv@v8.2.0 shape: ``src/download/download-version.ts``
+        # downloads via ``tc.downloadTool`` then calls ``validateChecksum(...)``,
+        # whose implementation was extracted into a sibling module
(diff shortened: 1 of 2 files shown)


dfoulks1 (approved):
LGTM

PR #918, opened 2026-06-08
Fix current Zizmor workflow failures

snazy:
1. `.github/workflows/codeql.yml` uses the "full" version
2. Let `gateway.py` emit a zizmor ignore rule for the 1password action


potiuk (approved):

potiuk:
Yep. quick fix.

PR #951, opened 2026-06-21
verify: exempt gradle-wrapper.jar from in-tree binary check

potiuk:
## What

Exempt `gradle/wrapper/gradle-wrapper.jar` from the in-tree binary check.

## Why

The in-tree binary check scans the whole action repo for pre-compiled native binaries and rejects any without verifiable provenance. It false-flagged `gradle/wrapper/gradle-wrapper.jar` on `JetBrains/qodana-action` (surfaced while triaging the 2026.1.0 → 2026.1.3 bump in #944).

That jar is **Gradle's build wrapper** — build-time tooling of the upstream repo, invoked by `./gradlew` during its own build, **never executed on a consumer's runner when the action runs** (qodana-action is a `node24` action that runs `dist/index.js`). It is also checksum-verifiable in its own right against Gradle's published list (cf. `gradle/wrapper-validation-action`). Flagging it as opaque action-runtime code is a false positive.

## How

- Add `_IN_TREE_BINARY_EXEMPT_PATH_SUFFIXES`, matched on the **canonical** `gradle/wrapper/gradle-wrapper.jar` path suffix (not bare name), so a same-named blob dropped anywhere else is still caught.
- Regression test covering: canonical path exempt, nested sub-path exempt, stray `dist/gradle-wrapper.jar` still flagged, generic `.jar` still flagged.

`uv run pytest utils/tests/verify_action_build/test_security.py` → 147 passed. prek clean.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -1405,6 +1405,22 @@ class TestLooksLikeInTreeBinary:
         assert _looks_like_in_tree_binary("dist/licenses.txt") is False
         assert _looks_like_in_tree_binary("licenses.txt") is False
 
+    def test_gradle_wrapper_jar_exempt(self):
+        # gradle/wrapper/gradle-wrapper.jar is Gradle's build wrapper:
+        # build-time tooling of the upstream repo, never executed on a
+        # consumer's runner, and checksum-verifiable in its own right.
+        # JetBrains/qodana-action (a node24 action) ships it and was
+        # false-flagged before this exemption.
+        assert _looks_like_in_tree_binary("gradle/wrapper/gradle-wrapper.jar") is False
+        # Exempt under a nested action sub-path too.
+        assert _looks_like_in_tree_binary(
+            "subdir/gradle/wrapper/gradle-wrapper.jar"
+        ) is False
+        # Precision: a stray jar of the same name elsewhere is still caught,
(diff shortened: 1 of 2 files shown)


snazy:
Hm, shouldn't projects not add binary executables to their source trees?

Making the Gradle build not require the jar to be present is pretty easy to achieve, I've documented a [possible approach](https://github.com/snazy/gradle-wrapper-no-jar). TL;DR is:
* add a human-reviewable [shell script snippet](https://github.com/apache/polaris/blob/main/gradle/gradlew-include.sh)
* add [one line](https://github.com/apache/polaris/blob/e3f26cf69f6ee189328f1b7adb90ec83cbbb26d9/gradlew#L91) to `gradlew`
* update the [root build-script](https://github.com/apache/polaris/blob/e3f26cf69f6ee189328f1b7adb90ec83cbbb26d9/build.gradle.kts#L266-L279) to get the `gradlew` properly re-generated

potiuk:
I guess it's not something we can "force" on others

snazy:
True, and it's what I'm going to. Maybe we can tell that it's better to have no executable binaries in the source tree?

potiuk:
Gentle ping @dfoulks1 @ppkarwasz — when you have a moment, this is a small, self-contained fix: it exempts Gradle's build-wrapper jar from the in-tree binary check (with a regression test). It also unblocks #944. Thanks!

dfoulks1 (approved):
LGTM

PR #970, opened 2026-06-27
verify: verify vendored node_modules against the npm registry

potiuk:
## What

Adds a deterministic, registry-anchored verification of vendored `node_modules` to `verify-action-build`, as a new **Vendored npm registry check**.

## Why

Node actions that commit their entire `node_modules` tree were only checked by an `npm ci` rebuild + file diff. npm isn't byte-reproducible across versions, so the rebuild frequently doesn't match and the tool falls back to a weak *"diff against the previously-approved version"* prompt — which forces a manual eyeball every bump. Surfaced reviewing **dawidd6/action-send-mail v17→v18** (#968), whose "DIFFERENCES DETECTED" was just a `nodemailer` 8→9 vendored refresh.

## How

For each package in the committed `node_modules/.package-lock.json` (lockfileVersion 2/3):

1. download the tarball from the lockfile `resolved` URL;
2. confirm its digest matches the lockfile `integrity` → proves it's the **published, untampered** package;
3. compare every file the tarball contains against the committed blob **by git blob SHA** — so committed blobs are never downloaded, the repo tree already carries them;
4. flag any committed file inside a verified package directory that the tarball does **not** contain (injected code).

A pass means *every vendored package matched a registry-published, integrity-verified tarball, with no extra files.*

### Safety / edge cases
- **Git/file/link deps** (no `integrity`) → reported as **skipped**, never silently passed.
- **Truncated repo tree** (too large to enumerate) → refuses to claim a pass.
- **Non-npmjs registries** → flagged.
- **`.bin`/`.cache`/`.package-lock.json`** npm-generated artifacts → ignored.
- Tarballs cached by integrity digest to avoid re-downloads.
- Needs network access to `registry.npmjs.org` (available in CI; local runs need it allowlisted).

## Tests

New `test_npm_registry_verify.py` (13 cases) builds a real in-memory tarball and exercises the full path with only the network seams mocked: clean match, content mismatch, integrity mismatch, extra-file injection, `.bin` noise, git-dep skip, foreign registry, truncated tree, plus helper unit tests (git-blob-SHA known value, integrity matching, tarball prefix stripping).

`uv run pytest utils/tests/verify_action_build/` green; prek clean.

### DIFF
--- a/utils/tests/verify_action_build/test_npm_registry_verify.py
+++ b/utils/tests/verify_action_build/test_npm_registry_verify.py
@@ -1,0 +1,172 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+"""Tests for the vendored-node_modules npm registry verification."""
+
+import base64
+import hashlib
+import io
+import json
+import tarfile
+from unittest import mock
+
+from verify_action_build import npm_registry_verify as nrv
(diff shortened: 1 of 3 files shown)


potiuk:
Gentle ping @ppkarwasz @dfoulks1 -- friendly nudge on this one (verifies vendored `node_modules` against the npm registry). Open since Jun 27; no urgency, but a look whenever you get a chance would be great. The one open CodeQL note is a substring-sanitization nit in a test file -- happy to tweak it if you'd prefer, otherwise I read it as a false positive on test data.

potiuk:
@dfoulks1 @ppkarwasz gentle nudge on this one - all checks green, and it has been waiting since late June. A look whenever you have a moment would be much appreciated. Happy to walk through the approach if that helps.

dfoulks1 (approved):
LGTM

PR #960, opened 2026-06-25
Add JetBrains/qodana-action v2026.1.3

lukaszlenart:
And sets previous version expiration within 3 months

PR #971, opened 2026-06-27
docs(skill): record qodana gradle-wrapper.jar precedent (#944/#960)

potiuk:
## What

Adds one row to the `analyze-action-pr` skill's **Recent precedents** table, recording the `JetBrains/qodana-action@v2026.1.3` review (#944 / #960).

## Why

Per the skill's own "keep it better than you found it" note, each triaged PR should leave a precedent line anchored to a real PR/issue so future runs can cite it instead of re-deriving.

The qodana review surfaced a **case-E** gap: the in-tree binary check scans the whole repo and false-flagged the committed `gradle/wrapper/gradle-wrapper.jar`, which is Gradle build tooling (never executed on a consumer's runner) and is checksum-verifiable in its own right. The fix is PR #951 (path-suffix exemption); upstream was also nudged to drop the committed jar via JetBrains/qodana-action#605.

Docs-only, one line.


dfoulks1 (approved):
LGTM

PR #977, opened 2026-06-29
verify-action-build: print a green attestation note with assumptions

potiuk:
## What

When a downloader script (or an `action.yml` run block) verifies its download via `gh attestation verify`, `verify-action-build` already passed it as *"verification present in file"* — but gave no signal that the strongest mechanism, **GitHub build attestation**, was the one detected.

This adds an explicit **green** note naming attestation verification, plus a `dim` **Assumptions** block that spells out the static-scan limitations:

```
✓ scripts/install.sh: 1 download(s), verification present in file
  ✓ GitHub build attestation verification detected (`gh attestation verify`) in scripts/install.sh
  Assumptions: this is a static text match. It confirms an attestation-verify
  command is present in the same file as the download; it does NOT run the
  installer, does NOT confirm the verify call targets the artifact that gets
  downloaded, and does NOT detect fail-open behaviour (e.g. verification
  skipped/warned when `gh` is unavailable). Treat it as 'verification mechanism
  present', not 'verification proven to be enforced'.
```

## Why

Surfaced while triaging #975 (dependabot bump of `untitaker/hyperlink` to 0.3.2, whose installer added `gh attestation verify`). The scan correctly passed it, but a reviewer couldn't tell *which* mechanism satisfied the check, nor what the check did and did not prove. The note makes the strongest signal visible and its assumptions explicit.

## Scope

- Pass/fail classification is **unchanged** — this only adds a clearer, caveated green signal.
- Fires for both inline `action.yml` run blocks and referenced installer scripts.
- Covered by a regression test modelled on the `untitaker/hyperlink` installer shape.

Full suite: `147 passed`.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -274,6 +274,37 @@ class TestAnalyzeBinaryDownloads:
                 warnings, failures = analyze_binary_downloads("org", "repo", "a" * 40)
         assert failures == []
 
+    def test_gh_attestation_verify_passes_and_prints_note(self, capsys):
+        # Mirrors the untitaker/hyperlink shape: action.yml runs an installer
+        # script that downloads from GitHub releases and verifies the download
+        # via `gh attestation verify`.
+        action_yml = """\
+name: Test
+runs:
+  using: composite
+  steps:
+    - name: Install
+      shell: bash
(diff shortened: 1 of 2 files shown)


dfoulks1 (approved):
LGTM

PR #1001, opened 2026-07-02
verify-action-build: recognize fileSHA256()-style file-hash helpers as verification

potiuk:
## Problem

The `verify` job on #980 (opentofu/setup-opentofu v2.0.1 → v2.0.2) fails with:

> ✗ lib/setup-tofu.js: 1 unverified JS download(s) (no checksum/signature check in file)

But the action **does** verify. `lib/setup-tofu.js` downloads the CLI zip and then:

```js
const checksum = await fileSHA256(pathToCLIZip);
if (!checksums.includes(checksum)) {
  throw new Error(`Failed to validate OpenTofu CLI zip checksum.`);
}
```

`fileSHA256` is imported from the sibling `lib/util.js`, where the actual
`createHash('sha256')` body lives, and `resolveChecksums()` falls back to the
release's published `SHA256SUMS`. Within the downloader file the only
verification evidence is the `fileSHA256(...)` **call name** — which no
`_JS_VERIFICATION_PATTERN` matched. So this is a false positive (case E), not a
real gap.

This is the same shape as the astral-sh/setup-uv `validateChecksum()` fix
(#910/#916): the hashing implementation lives in a sibling module and the call
name is the surviving in-file evidence.

## Fix

Add one pattern recognizing a helper call whose name contains an explicit SHA
algorithm (`fileSHA256(`, `getSHA512(`, `sha256File(`, …). It stays tight enough
not to match `createHash('sha256')` (the paren precedes the algorithm there) or
non-SHA hashes.

Two regression tests: the bare helper-name family, and a faithful trim of the
real opentofu download→verify sequence.

## Verification

- `uv run pytest utils/tests/` → 294 passed
- `prek run` on the changed files → clean

Once merged, #980 (and future opentofu/setup-opentofu bumps) verify cleanly.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -1360,6 +1360,43 @@ await validateChecksum(checksum, downloadPath, arch, platform, version);
         ):
             assert self._has_verification(snippet) is True, snippet
 
+    def test_file_sha_helper_name_recognized(self):
+        # ``fileSHA256(path)`` shape: a helper whose name denotes hashing a
+        # file with an explicit SHA algorithm.  Several actions define the
+        # ``createHash`` body in a sibling util module, so the call name is
+        # the in-file verification evidence.
+        for func_name in (
+            "fileSHA256",
+            "fileSHA512",
+            "getSHA256",
+            "sha256File",
+        ):
+            content = f"const checksum = await {func_name}(pathToCLIZip)\n"
+            assert self._has_verification(content) is True, func_name
+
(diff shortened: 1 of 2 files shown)


potiuk:
Gentle ping @ppkarwasz @dfoulks1 -- small verify-action-build enhancement here (recognizes `fileSHA256()`-style file-hash helpers as a verification step, so actions using that pattern aren't false-flagged). Open since Jul 2; no rush, but a review whenever you have a moment would be appreciated. Happy to walk through it if useful.

dfoulks1 (approved):
LGTM

PR #1013, opened 2026-07-07
Add jdx/mise-action to the GitHub Actions allowlist

kevinjqliu:
Add `jdx/mise-action` to the allowlist:

- `jdx/mise-action@5228313ee0372e111a38da051671ca30fc5a96db`

Needed by: `apache/iceberg-rust`

PR #1014, opened 2026-07-07
analyze-action-pr: record PR #1013 (jdx/mise-action) precedent

potiuk:
Records a `Recent precedents` row in the `analyze-action-pr` skill for the clean node20 pass on `jdx/mise-action@5228313` (v3.6.3, added in #1013).

Two reusable points captured:

- The verifier does not flag a tool-installer's `curl | tar` fetch of its own release tarball ("no binary downloads detected"), so that shape is a soft note, not a blocker.
- The signature-vs-same-source-checksum reasoning behind the gentle upstream hardening ask (jdx/mise-action#547): a checksum fetched from the same release as the binary is integrity, not authenticity; real supply-chain protection needs a signature verified against a key pinned in the action's own source.

Docs/skill-only change; no code affected.


potiuk:
@dfoulks1 @ppkarwasz small one - it just records the #1013 (jdx/mise-action) precedent for the analyze-action-pr skill. All checks green. A quick review when convenient would be great, thanks!

dfoulks1 (approved):
LGTM

PR #1032, opened 2026-07-11
Add tag-based, per-action releases (closes #1007)

### LINKED ISSUE #1007: Tag-based releases for infrastructure-actions
Today, consumers of the actions track the `main` branch for updates.

For example:

```yaml
- uses: apache/infrastructure-actions/allowlist-check@4e9c961f587f72b170874b6f5cd4ac15f7f26eb8  # main
```

https://github.com/apache/iceberg-rust/blob/0b45d614af89be40b4d96c95ded227a38e55a606/.github/workflows/asf-allowlist-check.yml#L42

However, it is good practice to pin GH Actions to specific hashes. This is often recommended by [Zizmor](https://zizmor.sh/), a static analysis tool for GHA.

The problem is that as new commits are made to `main`, this causes drift where the pinned hash no longer matches the ref (`main`). It's also unclear if Dependabot will make updates to the dependency in this scenario.

This issue is a request for tag-based releases, such that we can pin the action against a specific 'release'.

For example, we'd like to have something like this:

```yaml
- uses: apache/infrastructure-actions/allowlist-check@4e9c961f587f72b170874b6f5cd4ac15f7f26eb8  # v1.2.3
```

This would allow tools like Dependabot to recognize what version we're using, check for updates based on the Git tags for the repo, and Zizmor would not warn on drift.

potiuk:
Consumers of this repo's actions could previously only track `@main`, which drifts from any pinned SHA and which Dependabot cannot bump (#1007). This adds path-prefixed, per-action releases.

Since it's a monorepo of actions, each action is tagged under its **leaf-directory prefix** with a moving major tag:

| Action | Tag | Pin as |
|---|---|---|
| `allowlist-check` | `allowlist-check/vX.Y.Z` | `.../allowlist-check@<sha>  # allowlist-check/v1.2.3` |
| `pelican` | `pelican/vX.Y.Z` | `.../pelican@<sha>  # pelican/v1.2.3` |
| `stash/save` | `save/vX.Y.Z` | `.../stash/save@<sha>  # save/v1.2.3` |
| `stash/restore` | `restore/vX.Y.Z` | `.../stash/restore@<sha>  # restore/v1.2.3` |

This leaf-prefix scheme is exactly what Dependabot's `github_actions` ecosystem learned to support in [dependabot/dependabot-core#11286](https://github.com/dependabot/dependabot-core/pull/11286), contributed for this repo, so downstream projects pin a SHA with a `# <prefix>/vX.Y.Z` comment and receive automatic bump PRs.

## How it works

- `.github/workflows/release-actions.yml` runs on push to `main`; `scripts/release_actions.py` maps changed files to the affected action(s), computes the next version from the newest `<prefix>/vX.Y.Z` tag (first release seeds `v1.0.0`), creates the tag, moves the major tag, and publishes a GitHub Release.
- Bump defaults to **patch**; raise with a `release:minor` / `release:major` label (or `[minor]`/`[major]` in the PR) and suppress with `release:skip`. A `stash/shared/` change releases both stash actions.
- `workflow_dispatch` cuts a manual/seed release.
- Pure logic is unit-tested (`scripts/test_release_actions.py`, 25 tests); see `RELEASING.md`.

Merging this PR cuts no release (it touches no action directory); the first tags come from the next action change or a manual `workflow_dispatch` seed.

### DIFF
--- a/.github/workflows/release-actions.yml
+++ b/.github/workflows/release-actions.yml
@@ -1,0 +1,107 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# Cuts path-prefixed, per-action releases so downstream projects can pin a
+# version (and Dependabot can propose bumps). On every push to main the
+# release_actions.py script works out which action(s) changed, computes the
+# next version from the newest existing <leaf>/vX.Y.Z tag, creates the tag,
+# moves the <leaf>/vX major tag and publishes a GitHub Release. The bump type
+# defaults to patch and can be raised per PR with a release:minor /
+# release:major label (or suppressed with release:skip). See RELEASING.md.
+name: Release actions
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "allowlist-check/**"
+      - "pelican/**"
+      - "stash/**"
+  workflow_dispatch:
+    inputs:
+      action:
+        description: "Tag prefix to release (allowlist-check, pelican, save, restore). Leave empty to auto-detect from the latest commit."
+        required: false
+        default: ""
+      bump:
+        description: "Version bump to apply"
+        required: true
+        type: choice
+        default: patch
+        options:
+          - patch
+          - minor
+          - major
+
+permissions:
+  contents: read
+
+# Serialize releases so concurrent pushes to main can't race on tag creation.
+concurrency:
+  group: release-actions
+  cancel-in-progress: false
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          # Credentials are persisted so the script can push tags; full
+          # history + tags are needed to diff the pushed range and to find
+          # the newest existing version tag per action.
+          persist-credentials: true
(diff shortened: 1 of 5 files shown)


potiuk:
Hey: @dfoulks1 @ppkarwasz @dave2wave @gopidesupavan -> proposed solution to the issue we had whether to host actions in separate repos - because of dependabot release handling.

Since Pavan implemented feature in dependabot o handle prefixed version - it's now entirely doable to do the prefix/vX.Y.Z

The solution proposed in this PR is largely self-maintainable:

1) We merge
2) We manually run "worfflow-dispatch" to released first version of actions (say 1.0.0 for all - they are all well tested)
3) any future change to an action will automatically bump patchlevel - so all we need to do to relase new version is to push a fix.
4) we can also add  release:minor, release:major in a PR to bump Y or X instead

I think it's pretty good solution for zizmor/dependabot combo.

ppkarwasz (approved):
LGTM

dave2wave:
When I use the pelican action it looks like this:

https://github.com/apache/tooling-docs/blob/main/.github/workflows/build-pelican.yml#L23

Will this PR require changes?

potiuk:
> When I use the pelican action it looks like this:
> 
> https://github.com/apache/tooling-docs/blob/main/.github/workflows/build-pelican.yml#L23
> 
> Will this PR require changes?

No. This one will stil work. We only add a possibility of using tags, which will allow automated updates via dependabot.

Currently when you do this:

```
- uses: apache/infrastructure-actions/pelican@59478b84702ba730f7aa43755cc95579a04ece8f
```

It stays like that - and dependabot will not upgrade it to newer version.

When this one is merged (and initial tags created) you should be able to do this:

```
- uses: apache/infrastructure-actions/pelican@59478b84702ba730f7aa43755cc95579a04ece8f  # pelican/v1.0.0
```

And whenever new version of Pelican action is released, dependabot will upgrade it to newer version. The "releases" in our case will happen automatically - every time we merge something in the action - say - dependency update or a fix, new patchlevel tag will be automatically set (say pelican/v1.0.1) and dependabot will update the action together with other actions it upgrades already.

potiuk:
Also - when merging a minor or major change to an action, you should be able to set label on the PR (release:minor, release:major), and when the PR gets merged, the version of the action that is being modified will be bumped accordingly. 

It's pretty much self-maintainable, the only thing to remember is when adding new features or breaking changes to an action is to set the label.

There is small risk that the versions will be bumped too often - but we can always update the logic - for example I think, it might be happening that the version is updated always when dependencies to build the action change, which **might** be too often, but I think we can easily adjust the automation after a week or two observation of what makes sense and what kind of changes make sense to bump the verison - it's something best observed in practice - and since initially almost no-one will use it this way - they need to set the tag - we will have a chance to fine-tune it.

dave2wave (approved):
LGTM

PR #1034, opened 2026-07-12
gateway: emit zizmor unpinned-tools ignore off the uses: line (closes #952)

### LINKED ISSUE #952: dependabot leaves a stale version comment on uses: lines carrying a trailing # zizmor: ignore[...]
### Problem

When dependabot bumps an action whose `uses:` line carries a trailing `# zizmor: ignore[...]` *after* the version comment, its comment-updater skips the `# vX.Y.Z` version comment — it bumps the pinned hash but leaves the version stale. zizmor's [`ref-version-mismatch`](https://docs.zizmor.sh/audits/#ref-version-mismatch) (medium) then fails the `Run zizmor 🌈` check.

### Concrete case

#932 (1Password/load-secrets-action 4.0.0 → 4.0.1). In `.github/actions/for-dependabot-triggered-reviews/action.yml`, the hash became v4.0.1 but the comment stayed `# v4.0.0`:

```
- uses: 1Password/load-secrets-action@3a12b0ab…  # v4.0.0  # zizmor: ignore[unpinned-tools] generated sentinel step is never executed
```

The sibling `/configure` line (no trailing ignore comment) *was* updated correctly. Fixed manually in #932, but it will recur.

### Impact

Every future bump of an allowlisted action whose gateway-generated sentinel step carries a `# zizmor: ignore[unpinned-tools]` suffix will red the zizmor check and need a manual comment fix.

### Possible fixes

1. Have the `update`/gateway tooling re-sync version comments after dependabot runs.
2. Emit the `# zizmor: ignore[...]` on its own line (or before the version comment) so dependabot's regex still updates the version.
3. Auto-apply zizmor's `ref-version-mismatch` autofix as a pre-merge step.

Surfaced while reviewing #932.

potiuk:
Fixes #952.

## Problem

Dependabot's `github_actions` comment-updater only rewrites a `# vX.Y.Z` comment when the version is the *trailing* token -- `file_updater.rb` deliberately skips comments with text after the version (`comment.end_with?(version)`). The gateway emitted the sentinel step's zizmor ignore right after the version on the `uses:` line:

```
- uses: 1Password/load-secrets-action@<sha>  # v4.0.0  # zizmor: ignore[unpinned-tools] ...
```

so on a bump Dependabot updated the hash but left `# v4.0.0` stale, and zizmor's `ref-version-mismatch` (medium) then failed CI -- fixed by hand in #932 and set to recur on every future bump of an ignored action.

## Fix

Emit the ignore on the `if: false` line instead:

```yaml
    - uses: 1Password/load-secrets-action@<sha>  # v4.0.1
      if: false  # zizmor: ignore[unpinned-tools] generated sentinel step is never executed
      with:
        version: "2.30.0"
```

It's still within the sentinel step's finding span ([zizmor docs](https://docs.zizmor.sh/usage/#ignoring-results)), so `unpinned-tools` stays suppressed, while the version tag is now the trailing token on the `uses:` line, so Dependabot keeps it in sync.

- `gateway.py`: build the `uses:` comment from the tag only; append the ignore to the `if: false` line.
- Regenerated `.github/actions/for-dependabot-triggered-reviews/action.yml` (2-line diff).
- `test_gateway.py`: assert the new placement and that the ignore never lands on the `uses:` line.
- Verified: `gateway` / `allowlist-check` tests pass, prek clean, zizmor clean on the composite, and the composite -> `actions.yml` read-back round-trips with no diff.

### DIFF
--- a/gateway/test_gateway.py
+++ b/gateway/test_gateway.py
@@ -223,10 +223,18 @@ def test_generate_composite_action_includes_zizmor_ignore_only_for_selected_acti
 
     generated = generate_composite_action(actions)
 
+    # The version tag stays the trailing token on the `uses:` line (so
+    # Dependabot keeps it in sync); the zizmor ignore moves to the `if: false`
+    # line, within the sentinel step's finding span. See #952.
     assert (
-        f"1Password/load-secrets-action@92467eb28f72e8255933372f1e0707c567ce2259  "
-        f"# v4.0.0  # {ZIZMOR_UNPINNED_TOOLS_IGNORE}"
+        "1Password/load-secrets-action@92467eb28f72e8255933372f1e0707c567ce2259  "
+        "# v4.0.0\n"
     ) in generated
+    assert f"      if: false  # {ZIZMOR_UNPINNED_TOOLS_IGNORE}\n" in generated
+    # ... and never on the uses: line, which would break Dependabot's updater.
+    for line in generated.splitlines():
+        if "1Password/load-secrets-action@92467" in line:
+            assert "zizmor: ignore" not in line
     assert (
         "DavidAnson/markdownlint-cli2-action@ded1f9488f68a970bc66ea5619e13e9b52e601cd  "
         "# v23.2.0"
--- a/.github/actions/for-dependabot-triggered-reviews/action.yml
+++ b/.github/actions/for-dependabot-triggered-reviews/action.yml
@@ -37,8 +37,8 @@ name: Gateway Action
 runs:
   using: "composite"
   steps:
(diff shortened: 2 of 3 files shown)


raboof (approved):
workaround seems fine.

gateway/test_gateway.py
```diff
@@ -223,10 +223,18 @@ def test_generate_composite_action_includes_zizmor_ignore_only_for_selected_acti
 
     generated = generate_composite_action(actions)
 
+    # The version tag stays the trailing token on the `uses:` line (so
+    # Dependabot keeps it in sync); the zizmor ignore moves to the `if: false`
+    # line, within the sentinel step's finding span. See #952.
     assert (
```
raboof, line 229: ideally comments should describe the current state, not the last change.

Issue #891, opened 2026-05-27
maintain-one-comment should be removed from allowed list due to being compromised and no longer available


rluvaton:
[maintain-one-comment](https://github.com/actions-cool/maintain-one-comment)

See more [here](https://www.stepsecurity.io/blog/actions-cool-issues-helper-github-action-compromised-all-tags-point-to-imposter-commit-that-exfiltrates-ci-cd-credentials)

Issue #666, opened 2026-04-05
Review and where possible remove wildcard/keep actions


raboof:
> Please go ahead and remove all wildcard marketplace actions so that they may be replaced as needed in the future upon request. 

 _Originally posted by @gmcdonald in [#252](https://github.com/apache/infrastructure-actions/issues/252#issuecomment-3967952942)_

dave2wave:
Here are all the actions approved with '*' / keep

- [ ] actions-cool/check-user-permission:
- [ ] actions-cool/issues-helper:
- [ ] actions-cool/maintain-one-comment:
- [ ] AdoptOpenJDK/install-jdk:
- [ ] amannn/action-semantic-pull-request:
- [ ] arduino/setup-protoc:
- [ ] awalsh128/cache-apt-pkgs-action:
- [ ] BobAnkh/auto-generate-changelog:
- [ ] bufbuild/buf-breaking-action:
- [ ] bufbuild/buf-lint-action:
- [ ] bufbuild/buf-setup-action:
- [ ] burrunan/gradle-cache-action:
- [ ] bytedeco/javacpp-presets:
- [ ] carlosperate/arm-none-eabi-gcc-action:
- [ ] check-spelling/check-spelling:
- [ ] chromaui/action:
- [ ] cicirello/javadoc-cleanup:
- [ ] clechasseur/rs-cargo:
- [ ] codecov/codecov-action:
- [ ] codelytv/pr-size-labeler:
- [ ] codspeedhq/action:
- [ ] conda-incubator/setup-miniconda:
- [ ] container-tools/kind-action:
- [ ] container-tools/microshift-action:
- [ ] coursier/setup-action:
- [ ] crate-ci/typos:
- [ ] dawidd6/action-download-artifact:
- [ ] delaguardo/setup-graalvm:
- [ ] dlang-community/setup-dlang:
- [ ] easimon/maximize-build-space:
- [ ] EnricoMi/publish-unit-test-result-action:
- [ ] eps1lon/actions-label-merge-conflict:
- [ ] geekyeggo/delete-artifact:
- [ ] golang/govulncheck-action:
- [ ] golangci/*:
- [ ] gr2m/twitter-together:
- [ ] gsactions/commit-message-checker:
- [ ] houseabsolute/actions-rust-cross:
- [ ] ilammy/msvc-dev-cmd:
- [ ] jarvusinnovations/background-action:
- [ ] jidicula/clang-format-action:
[...]

PR #1036, opened 2026-07-12
Remove 14 unused wildcard actions from allowlist (part of #666)

potiuk:
Part of #666, motivated by #891.

Each of these actions is allowlisted with a wildcard (`'*': keep: true`) but is **not referenced by any `apache/*` repository** (checked via org-wide code search on default branches). A wildcard approves *any* ref of an action, which -- as the actions-cool compromise (#891) showed -- is the exact door a hijacked tag walks through. Removing unused wildcards shrinks that attack surface and keeps the list under its size cap; any of these can be re-added on demand (with review) if a project actually needs it.

### Removed (14)

```
bytedeco/javacpp-presets           jarvusinnovations/background-action
check-spelling/check-spelling      mvasigh/dispatch-action
clechasseur/rs-cargo               shufo/auto-assign-reviewer-by-files
codspeedhq/action                  timonvs/pr-labeler-action
container-tools/microshift-action  wei/curl
delaguardo/setup-graalvm           xpol/setup-lua
golang/govulncheck-action          gsactions/commit-message-checker
```

`approved_patterns.yml` updated to match (surgical diff).

### Not touched

- **4 namespace-wildcards** (`golangci/*`, `r-lib/actions/*`, `quarto-dev/quarto-actions/*`, `rustsec/*`) are in active use -- kept, flagged in #666 as candidates to narrow to specific sub-actions in a follow-up.
- **83 single-action wildcards** in active use are kept.

### Caveat

"Unused" = no usage found on indexed default branches; a usage on a non-default branch or an un-indexed repo would not appear. Treated as strong candidates, not blind deletions. Full analysis in #666.

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -157,9 +157,6 @@ burnett01/rsync-deployments:
 burrunan/gradle-cache-action:
   '*':
     keep: true
-bytedeco/javacpp-presets:
-  '*':
-    keep: true
 carabiner-dev/actions/ampel/verify:
   e0e3b8149dafed833431095bc148d50e7eade4e8:
     tag: v1.2.0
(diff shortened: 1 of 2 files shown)


raboof (approved):
awesome!

typically we ask that you only edit `actions.yml` and leave `approved_patterns.yml` to be automatically updated, but I guess it can't hurt.

PR #1038, opened 2026-07-13
allowlist-check: warn about upcoming action expirations (closes #987)

### LINKED ISSUE #987: allowlist-check: warn about upcoming expirations
Feature request for `allowlist-check`: produce [warning annotations](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands?versionId=free-pro-team%40latest&productId=actions#setting-a-warning-message) for any actions that are currently allowed but have expiration in the near future.

Example:

https://github.com/apache/infrastructure-actions/blob/775350a154e610e84c460cb1bbe2d2ab26c15cb3/actions.yml#L72-L75

CC @kevinjqliu @potiuk

potiuk:
Closes #987.

## Problem

Approved action *versions* don't live forever: when a newer version is approved, an older pinned SHA is given an `expires_at` grace period (typically ~3 months) and is eventually removed from the allowlist. Once removed, any workflow still pinning that SHA starts failing the allowlist check -- with no prior warning.

## What this adds

`allowlist-check` now also fetches [`actions.yml`](../actions.yml) (which carries the `expires_at` metadata that `approved_patterns.yml` doesn't) and emits a **non-failing** `::warning::` for any allowlisted pin in the caller's workflows that expires within `expiry-warning-days` (new input, default **30**):

```
1 allowlisted ref(s) expiring within 30 day(s) -- bump these to a newer approved version before they are removed:
::warning file=.github/workflows/ci.yml::some-org/some-action@abc123 is allowlisted but its approved pin expires on 2026-07-25 (in 12 day(s)); bump to a newer approved version to avoid a future CI failure.
```

Warnings never change the exit code -- they surface in the job log and PR annotations so projects get advance notice.

## Details

- `check_asf_allowlist.py`: `load_expiry_map()` + `upcoming_expiry_warnings()` (pure, unit-tested) helpers, `emit_expiry_warnings()`, and an **optional** second CLI arg for `actions.yml`. The 1-arg invocation stays fully backward-compatible (no warnings, unchanged behavior).
- `action.yml`: fetches `actions.yml` best-effort (if it can't be fetched, the check still runs, just without expiry warnings) and passes the new `EXPIRY_WARNING_DAYS`.
- New tests (48 pass total) + README docs (input table + "Expiry warnings" section).

Set to `0` to warn only once a pin has actually expired, or raise it (e.g. `"60"`) for earlier notice.

## Note

I implemented this from the issue title/intent -- GitHub's API was unreachable from my environment when I built it, so I couldn't read the issue body verbatim. If #987 specified a different threshold default, warn-vs-fail behavior, or data source, say the word and I'll adjust.

### DIFF
--- a/allowlist-check/README.md
+++ b/allowlist-check/README.md
@@ -63,6 +63,7 @@ That's it — two steps. The `actions/checkout` step checks out your repo so `.g
 | Input | Required | Default | Description |
 |---|---|---|---|
 | `scan-glob` | No | `.github/**/*.yml` | Glob pattern for YAML files to scan for action refs |
+| `expiry-warning-days` | No | `30` | Emit a non-failing warning when a workflow pins an allowlisted action whose approved version expires within this many days. Set to `0` to warn only once a pin has actually expired. |
 
 ### Custom scan glob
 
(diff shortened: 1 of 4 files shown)


raboof (approved):

raboof:
This should ideally rarely happen since dependabot (or equivalent) should be creating update PRs and these should be reviewed and merged, but I can see how it might be helpful to have an additional signal that such an update is still pending.

adoroszlai:
Thanks a lot!

PR #1045, opened 2026-07-14
Update blog card date formatting

Paul-TT:
Adding helper method to format blog rss feed date to be more human friendly.

### DIFF
--- a/pelican/plugins/asfdata.py
+++ b/pelican/plugins/asfdata.py
@@ -561,6 +561,29 @@ def truncate_words(text, words):
     return content_text
 
 
+def format_blog_date(date_text):
+    if not date_text:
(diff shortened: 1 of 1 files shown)


potiuk (approved):
Approving. Verified `import datetime` is present so `datetime.datetime.*` resolves correctly and matches existing usage in this module; the helper degrades gracefully (empty string on empty input, original text when unparseable) and uses `.day` to avoid zero-padding.

Two minor non-blocking notes: `strftime("%B")` is locale-dependent, and `replace('Z', '+00:00')` replaces all occurrences rather than just a trailing Z - neither matters for the ISO-8601 timestamps this feed produces.

Thanks for the contribution!

PR #1106, opened 2026-07-27
release-actions: report GitHub Release failures instead of swallowing them

potiuk:
Found while seeding the first tags via `workflow_dispatch` (run [30281213015](https://github.com/apache/infrastructure-actions/actions/runs/30281213015)): `allowlist-check/v1.0.0` and `allowlist-check/v1` were pushed, **no GitHub Release was created, and the run reported success**.

- `create_release()` published through `_run(..., check=False)`, which pipes stderr and discards it on a non-zero exit. A failing `gh release create` therefore printed nothing, the script continued and exited 0.
- Publishing now goes through `_run_reporting()`, which prints the exit code and stderr as an `::error::` annotation and returns the outcome.
- `main()` collects the actions whose Release did not publish and exits 1, so the run goes red instead of green.
- Tags are still pushed when the Release fails - they are what consumers and Dependabot pin against - but the gap is now reported rather than silent.

The underlying cause of the `gh` failure is still open (most likely `ALLOWLIST_WORKFLOW_TOKEN` lacking permission to create releases). This PR does not fix that; it makes the next run print the actual API error instead of hiding it.

## Test plan

- 8 new tests in `scripts/test_release_actions.py` covering: stderr surfaced on failure, the empty-output case, `create_release` still pushing tags when the Release fails, and `main()` exit codes both ways.
- `uv run pytest scripts/ utils/tests/` - 340 passed.

To verify against the real failure, dispatch `Release actions` with `--ref fix/release-surface-gh-errors`; the run should now fail with the actual `gh` error annotated.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -20,6 +20,10 @@ name: Linting and MyPy (Pelican)
 
 on:
   push:
+    # Branches only: release tags (pelican/v1.0.0, ...) touch these paths and
+    # re-running the linters on a tag adds nothing over the branch run.
+    branches:
+      - '**'
     paths:
       - 'pelican/**.py'
       - 'pelican/pyproject.toml'
--- a/.github/workflows/pelican-action-test.yml
+++ b/.github/workflows/pelican-action-test.yml
@@ -19,6 +19,12 @@
 name: Test the Pelican build action
 on:
   push:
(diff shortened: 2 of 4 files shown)


potiuk:
Closing: this was pushed to test the release-tag seeding against real CI, and the seeding is being done by creating the tags directly instead. The branch is kept so the two commits can be re-landed if wanted.

PR #1125, opened 2026-07-30
verify-action-build: treat only-in-rebuilt JS as informational

potiuk:
The JS build check hard-failed on any file present in the rebuild but absent
from the published tree. Such a file is not published by the action, so it
never reaches a consumer's runner and cannot be a supply-chain vector -- it is
normally intermediate output from a multi-stage build that upstream
deliberately does not commit.

`JetBrains/qodana-action` is the recurring case. It declares
`main: scan/dist/index.js`, so `OUT_DIR` resolves to the whole `scan/`
sub-project (the Dockerfile takes the first path component of `main:`), and the
rebuild's stage-one `tsc` output `scan/lib/*.js` lands inside the compared
tree. That output is gitignored upstream via `**/lib`, is never committed, and
is consumed only by the `esbuild` step that writes the published
`scan/dist/index.js`.

This failed the check on #960 -- the PR that approved v2026.1.3, merged over
the failure anyway -- and again now on #1123. Only the gradle-jar half of
#960's failure was ever fixed (in #951). Each run also triggers the
approved-lock-file retry rebuild, which re-emits the same files and fails
identically, burning ~3 minutes of CI that can never help for this shape.

## What changes

Only-in-rebuilt files are now reported rather than failed on.

What actually matters is unchanged:

- JS in the published tree that **differs** from the rebuild: still a hard
  failure.
- JS in the published tree that the rebuild **does not produce**: still a hard
  failure (unaccounted-for shipped code).
- New guard for the one case the downgrade could have masked: a published tree
  with **no** compiled JS while the rebuild produced some, where there is
  nothing to reconcile the rebuild against.

## Tests

Four regression tests using qodana-action's real shape (published
`scan/dist/index.js` + `scan/jest.config.js`; rebuild adds
`scan/lib/{annotations,main,output,utils}.js`), covering the false positive and
each preserved failure mode. Full suite: 311 passed. `prek run --all-files`
clean.

README's "Verifying Compiled JavaScript" section documents the new behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,8 @@ A clean result confirms that the compiled JS was built from the declared source.
 
 Non-minified compiled JS (e.g. Deno `deno task bundle` output, Dart `dart compile js` readable output) is handled differently: a clean rebuild for these tends to produce toolchain-version noise (esbuild/ncc/webpack boilerplate differences) rather than actionable diffs. The script keeps these files in place during the pre-rebuild deletion step and instead diffs them against the previously approved version of the action, so reviewers see real source changes rather than rebuild artifacts. The detection threshold mirrors the comparison heuristic — fewer than 10 lines or an average line length above 500 chars is treated as minified.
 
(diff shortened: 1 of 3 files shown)


potiuk:
Context for reviewers: the JS build check hard-failed on files present in the rebuild but missing from the published tree. Those are never shipped, so they can't run on a consumer's runner. The recurring case is `JetBrains/qodana-action`: its two-stage `tsc` -> `esbuild` build leaves gitignored `scan/lib/*.js` in the compared tree, which failed #960 (merged over anyway) and #1123. What matters is unchanged -- published JS that differs from the rebuild, or that the rebuild doesn't produce, still hard-fails.

dave2wave (approved):
From a glance this looks ok.

PR #1135, opened 2026-08-03
analyze-action-pr: document gh keyring and docker PATH gotchas

potiuk:
Follow-up to the #1133 / #1134 triage. Two environment failures in
`verify_action_build` report causes that point away from the real problem, and
both cost a diagnosis round-trip:

- A keyring-stored `gh` token is unreachable from the child process the CLI
  spawns, so the nested `gh pr diff` gets HTTP 401 and the CLI prints
  `could not fetch diff for PR #<N>` even though `gh auth status` reports a
  healthy login.
- Docker Desktop installs its CLI in `~/.docker/bin`, which may be off `PATH`
  (or a sandbox-denied read path), so `docker is required but not found in
  PATH` fires while the daemon is running fine.

Adds to the `analyze-action-pr` skill:

- An **Environment gotchas** table under Prerequisites with the symptom, cause
  and fix for each, plus the one-line test that identifies the keyring case
  (`bash -c 'gh auth status'` fails while a direct `gh auth status` succeeds).
- A **Recent precedents** row for #1133 / #1134: when a tag-pinned `FROM`
  warning is already present on the approved version, it is a standing
  property of the action rather than a regression the bump introduced.

Docs-only; no code touched. `prek run --all-files` clean.

### DIFF
--- a/.claude/skills/analyze-action-pr/SKILL.md
+++ b/.claude/skills/analyze-action-pr/SKILL.md
@@ -35,6 +35,20 @@ before sending.
 - `uv` installed (the verify-action-build CLI is invoked through it).
 - For node-action JS rebuilds: Docker running.
 
+### Environment gotchas
+
+`verify_action_build` shells out to `gh` and `docker`, so it needs both to
+work *from a child process* — not just from your own shell. Two setups break
+that, and both surface as errors that point away from the real cause:
+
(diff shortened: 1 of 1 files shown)


dfoulks1 (approved):
This also LGTM. is this ready to merge?

PR #1100, opened 2026-07-25
Pin `reactivecircus/android-emulator-runner`

ppkarwasz:
- Replaces the `reactivecircus/android-emulator-runner@*` wildcard with the v2.38.0 SHA (`a421e438`) as the permanent, Dependabot-maintained pin.
- Whitelists the v2.37.0 and v2.34.0 SHAs and the floating `v2` tag until 2026-08-25; `remove_expired.yml` will drop these automatically.
- Composite action and `approved_patterns.yml` regenerated with the gateway sync.

Test plan: `uv run pytest` passes (544 passed, 5 skipped).

## Repos to migrate before 2026-08-25

Using the floating `v2` tag (11 repos):

| Repo | Workflow |
|---|---|
| apache/fory | `.github/workflows/ci.yml` |
| apache/cordova-plugin-device | `.github/workflows/android.yml` |
| apache/cordova-plugin-device-motion | `.github/workflows/android.yml` |
| apache/cordova-plugin-device-orientation | `.github/workflows/android.yml` |
| apache/cordova-plugin-dialogs | `.github/workflows/android.yml` |
| apache/cordova-plugin-file | `.github/workflows/android.yml` |
| apache/cordova-plugin-geolocation | `.github/workflows/android.yml` |
| apache/cordova-plugin-media | `.github/workflows/android.yml` |
| apache/cordova-plugin-media-capture | `.github/workflows/android.yml` |
| apache/cordova-plugin-screen-orientation | `.github/workflows/android.yml` |
| apache/cordova-plugin-statusbar | `.github/workflows/android.yml` |

Pinned to v2.37.0 (`e89f39f1abbbd05b1113a29cf4db69e7540cae5a`):

| Repo | Workflow |
|---|---|
| apache/cordova-plugin-inappbrowser | `.github/workflows/android.yml` |
| apache/cordova-plugin-network-information | `.github/workflows/android.yml` |

Pinned to v2.34.0 (`1dcd0090116d15e7c562f8db72807de5e036a4ed`):

| Repo | Workflow |
|---|---|
| apache/cordova-plugin-vibration | `.github/workflows/android.yml` |

Already on the v2.38.0 SHA, no change needed: apache/commons-xml, apache/cordova-paramedic, apache/cordova-plugin-battery-status, apache/cordova-plugin-camera, apache/cordova-plugin-file-transfer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1136, opened 2026-08-03
verify-action-build: normalize package.json before comparing

potiuk:
Found while triaging #1100 (`reactivecircus/android-emulator-runner`), which
fails `verify` on four checks. Three are tooling gaps; this fixes them.

## package.json is rewritten at install time

A vendored `node_modules` holds the manifest **npm rewrote when installing**,
while the registry tarball and a fresh `npm ci` hold the **published** one.
Comparing bytes reports that difference as tampering. Two separate rewrites:

- **Install bookkeeping** — npm writes `_`-prefixed keys that the tarball never
  had. `diff_node_modules` already filtered these, but by enumeration, and the
  list was missing `_args` and `_location`. Now matched by the reserved `_`
  prefix, since the set has varied across npm versions.
- **Shorthand expansion** — `normalize-package-data` turns an `author` string
  into a person object, `bugs` into `{"url": ...}`, and prefixes
  `repository.url` with `git+`.

`node_modules/tunnel/package.json` hits both: it was the *only* differing file
out of 217, and on its own failed the whole JS build check and the registry
check.

Normalisation is deliberately narrow. Only the fields npm is known to rewrite
are touched; everything else — `main`, `bin`, `scripts`, `dependencies` — is
still compared strictly, with a regression test asserting a tampered `main`
still fails.

## Two smaller fixes

- **`@actions/tool-cache`'s `7zdec.exe`** is exempted from the in-tree binary
  check. It backs `tc.extractZip` on Windows and ships inside the first-party
  npm package, so every action vendoring `node_modules` carries it verbatim
  from the published tarball. Same narrow path-suffix form as the
  `gradle-wrapper.jar` exemption in #951. The general fix — crediting any
  binary under a `node_modules` package the registry check already verified —
  needs the two checks reordered, so I left it for a follow-up.
- **The registry mismatch line** passed the offending path as the *URL* and the
  tree root as the *text*, so it never named the failing file. Diagnosing #1100
  meant deriving the filename from the rebuild diff instead.

## Test plan

- `uv run pytest utils/tests/` — 320 passed.
- Normaliser checked against the real bytes: committed
  `node_modules/tunnel/package.json` at `a421e438` vs the published
  `tunnel-0.0.6.tgz` from the registry now compare equal.

Not yet re-run end-to-end against the action itself — that needs a
`GITHUB_TOKEN` and I exhausted the unauthenticated rate limit. CI on this PR
will cover it.

### DIFF
--- a/utils/verify_action_build/security.py
+++ b/utils/verify_action_build/security.py
@@ -1789,6 +1789,16 @@ _IN_TREE_BINARY_EXEMPT_PATH_SUFFIXES = (
     # gradle/wrapper-validation-action).  Seen on JetBrains/qodana-action,
     # a node24 action that never shells out to Gradle at runtime.
     "gradle/wrapper/gradle-wrapper.jar",
+    # 7-Zip's standalone decoder, bundled inside the first-party
+    # @actions/tool-cache npm package (it backs ``tc.extractZip`` on
+    # Windows).  It is not the action's own code: any action that vendors
+    # its node_modules ships it verbatim from the published tarball, and
+    # the Vendored npm Registry Check verifies that package's bytes against
+    # the lockfile integrity hash.  Seen on reactivecircus/android-emulator-
+    # runner.  The general fix — crediting any binary under a node_modules
+    # package the registry check already verified — needs the two checks
+    # reordered, so this keeps the narrow, reviewable form used above.
+    "node_modules/@actions/tool-cache/scripts/externals/7zdec.exe",
 )
 
 
--- a/utils/tests/verify_action_build/test_diff_node_modules.py
+++ b/utils/tests/verify_action_build/test_diff_node_modules.py
@@ -183,3 +183,66 @@ class TestDiffNodeModules:
             orig, rebuilt, "test", "repo", "a" * 40,
         )
         assert result is True
+
+    def test_package_json_npm6_install_metadata_ignored(self, tmp_path):
+        # Real shape from reactivecircus/android-emulator-runner@a421e438:
+        # node_modules/tunnel/package.json was installed with npm v6-era
+        # tooling, which writes _args/_location (among others) that a modern
+        # npm ci rebuild does not.  An enumerated field list missed exactly
+        # those two and reported the package as modified — the only "content
+        # difference" in 217 files, which failed the whole JS build check.
+        import json
+        orig = tmp_path / "original"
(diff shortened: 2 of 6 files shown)


potiuk:
End-to-end validation done - this closes the caveat in the PR description.

Ran the verifier from this branch against `reactivecircus/android-emulator-runner@a421e438` with a token, so no rate-limit degradation:

| Check | Before | After |
| --- | --- | --- |
| Vendored npm registry check | FAIL - 1 modified | PASS - 14 packages match registry-verified tarballs |
| JS build verification | FAIL - differences detected | PASS - compiled JS matches rebuild |
| In-tree binary check | FAIL - unverified 7zdec.exe | PASS |
| Binary download verification | FAIL - 2 unverified | FAIL - 2 unverified (unchanged, expected) |

The remaining failure is the unchecksummed cmdline-tools fetch from dl.google.com. That is the action's own behaviour, now tracked upstream at ReactiveCircus/android-emulator-runner#485. Exit code stays 1 on that single finding, which is correct.

Note that `tunnel` moves into the verified set rather than merely being tolerated: the registry check reports 14 verified packages instead of 13 plus one modified.

`uv run pytest utils/tests/` - 320 passed.

---
Drafted-by: Claude Opus 5 (1M context) via Claude Code; reviewed by @potiuk before posting

utils/verify_action_build/security.py
```diff
@@ -1789,6 +1789,16 @@ def analyze_repo_metadata(
...
+    # 7-Zip's standalone decoder, bundled inside the first-party
+    # @actions/tool-cache npm package (it backs ``tc.extractZip`` on
+    # Windows).  It is not the action's own code: any action that vendors
+    # its node_modules ships it verbatim from the published tarball, and
+    # the Vendored npm Registry Check verifies that package's bytes against
+    # the lockfile integrity hash.  Seen on reactivecircus/android-emulator-
+    # runner.  The general fix — crediting any binary under a node_modules
+    # package the registry check already verified — needs the two checks
+    # reordered, so this keeps the narrow, reviewable form used above.
+    "node_modules/@actions/tool-cache/scripts/externals/7zdec.exe",
```
ppkarwasz, line 1801: We could verify its SHA256, because here we have the same binary since 10 years.

dfoulks1 (approved):
LGTM

Issue #1153, opened 2026-08-06
how to pin `dtolnay/rust-toolchain`


raboof:
dtolnay/rust-toolchain does not use tags: the commonly-used `stable` but also versions like `1.89.0` are not tags, but branches. https://github.com/dtolnay/rust-toolchain#choice-of-full-length-commit-sha recommends picking commits that are part of the master branch, warning other commits will be garbage-collected - so presumably neither the 'channel branches' nor the 'version branches' are expected to be usable longer-term (?). Perhaps we should track the latest commits to `master` of the action? (the action has a parameter to select which toolchain to install, so that might work fine).

Alternatively, perhaps we could recommend another method of installing a Rust toolchain? Perhaps https://github.com/actions-rust-lang/setup-rust-toolchain ?

/cc @snazy @potiuk @kevinjqliu

hainenber:
+1 to `actions-rust-lang/setup-rust-toolchain`. It's well-maintained, pinnable and augmented with many options.

PR #1167, opened 2026-08-09
Allow jdx/mise-action v4.2.4

tisonkun:

## Overview

**Name of action:** `jdx/mise-action`

**URL of action:** https://github.com/jdx/mise-action

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):** `7e36c90d9ab29c415a2384db3006f3ec8a8cc654` (`v4.2.4`)

Apache DataSketches TCK is migrating its tool setup and checks to mise in apache/datasketches-tck#3. The currently approved `v3.6.3` release runs on Node.js 20, which GitHub Actions now warns is deprecated. `v4.2.4` runs on Node.js 24 and includes the latest PATH propagation and lock-detection fixes. The existing approved SHA is retained for current consumers.

## Permissions

The action uses the automatic GitHub token for authenticated release API requests and reads or writes the GitHub Actions cache by default. It does not require repository write access. Its optional OIDC-based cache integration is disabled by default and is not used by DataSketches TCK.

## Related Actions

`jdx/mise-action@5228313ee0372e111a38da051671ca30fc5a96db` (`v3.6.3`) is already approved. This request adds the current v4 release rather than replacing the existing SHA.

## Checklist

- [ ] The action is not already on the list of approved actions (v3.6.3 is already approved)
- [x] Compiled JavaScript in `dist/` matches a clean rebuild (verify with `uv run utils/verify-action-build.py org/repo@hash`)

## Verification note

`uv run utils/verify-action-build.py --ci jdx/mise-action@7e36c90d9ab29c415a2384db3006f3ec8a8cc654` confirms Node.js 24, signed-checksum download verification, and that the compiled JavaScript matches a clean rebuild. The command currently exits non-zero on lock-file presence because v4.2.4 uses `aube-lock.yaml`, which the verifier does not yet recognize. This PR is draft pending guidance on adding aube support to the verifier.

---
Drafted-by: Codex (GPT-5)

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -686,6 +686,8 @@ jasonetco/create-an-issue:
 jdx/mise-action:
   5228313ee0372e111a38da051671ca30fc5a96db:
     keep: true
+  7e36c90d9ab29c415a2384db3006f3ec8a8cc654:
+    tag: v4.2.4
 JetBrains/qodana-action:
   d7b5ec2fbec32197ef447c450e00589ed5f34fd5:
     tag: v2026.1.0
--- a/actions.yml
+++ b/actions.yml
@@ -629,6 +629,8 @@ jasonetco/create-an-issue:
 jdx/mise-action:
   5228313ee0372e111a38da051671ca30fc5a96db:
     keep: true
+  7e36c90d9ab29c415a2384db3006f3ec8a8cc654:
+    tag: v4.2.4
 JetBrains/qodana-action:
   d7b5ec2fbec32197ef447c450e00589ed5f34fd5:
     tag: v2026.1.0


raboof:
`check_action_tags` failure is #1153, can be ignored here I think.

'Verify Manual Allowlist PR` is not recognizing the [aube-lock.yaml](https://github.com/jdx/mise-action/blob/main/aube-lock.yaml) as a lockfile. Probably fine but might be good if someone with more nodejs experience had a closer look?

potiuk:
The block here is our tooling, not your PR. `verify-action-build`'s lock-file check only knew `package-lock.json`, `yarn.lock`, `pnpm-lock.yaml`, `bun.lock` and `bun.lockb` - mise-action v4.2.4 moved its build to aube, which writes `aube-lock.yaml`. That file is pnpm lockfileVersion 9 with resolved versions and sha512 integrity for all 460 transitive deps, so the pinning guarantee the check exists to enforce is fully met.

Fix is up in #1175 (teaches the check the filename, and teaches the rebuild container to install from it). Once that merges I'll re-run CI here and this should go green. Nothing needed from you.

---

potiuk:
@raboof - following up on your node question above, both halves now have fixes up and green:

- **#1175** teaches the lock-file check about `aube-lock.yaml`. It is pnpm lockfileVersion 9 with resolved versions plus sha512 integrity for all 460 transitive deps, so the pinning guarantee the check exists to enforce is fully met. It also teaches the rebuild to install from it, which otherwise fell through to an unpinned `npm install`. Verified pnpm 11 consumes the real file in a `node:24-slim` container.
- **#1174** fixes a separate false positive in the same tool, surfaced by #1171: npm tarball roots were assumed to be `package/`, so DefinitelyTyped `@types/*` tarballs had every file reported as injected code after already passing integrity verification.

Neither is a change to what gets enforced, only to the tool recognising what is in front of it. Once they land, this PR and #1171 just need a branch update to re-run `verify`. Would you take a look?

@tisonkun - nothing needed from you here, the block was on our side.

potiuk:
The `verify` failure here is a gap in our own tooling, not a problem with the action.

The single red check is `Lock file presence`. Everything else is green, including `JS build verification` (compiled JS matches the rebuild). mise-action v4.2.4 moved its build to [aube](https://github.com/jdx/aube), which writes `aube-lock.yaml` instead of `package-lock.json`. That file is pnpm lockfileVersion 9 format and pins all 460 transitive deps with sha512 integrity, so the guarantee is the same one the other lock files give. Our presence check only knew the five established filenames.

#1175 teaches the check (and the rebuild) about `aube-lock.yaml`. Once it lands, re-running `verify` here should go green.

@dfoulks1 @ppkarwasz - #1175 is waiting on a review; this PR is blocked behind it.

---

potiuk (approved):
Green after rebase onto current main.

This was held on a verify-action-build gap, not on the action. v4.2.4 moved the build to aube, whose aube-lock.yaml the lock-file presence check did not recognise, so a repo that does have a lock file was reported as missing one. Fixed in #1175, and verify now passes clean.

PR #1174, opened 2026-08-12
verify-action-build: detect the tarball root instead of assuming package/

potiuk:
# Code change

## Summary

`_tarball_files()` stripped a literal `package/` prefix from npm tarball members. DefinitelyTyped roots `@types/*` tarballs at the bare package name instead (`estree/index.d.ts`, `json-schema/LICENSE`), so no extracted path lined up with `node_modules/@types/<pkg>/<rel>`, nothing was accounted for, and every file of every vendored `@types` package was reported as injected code - after that tarball had already passed integrity verification.

The root directory is now detected rather than assumed, and `./`-prefixed member names are normalised. A tarball whose files do not all share one root is left untouched.

Surfaced by #1171 (`JamesIves/github-pages-deploy-action` v4.9.0), whose yarn -> npm migration added `node_modules/.package-lock.json` and so switched this check on for the first time. It reported 13 extra files across `@types/esrecurse`, `@types/estree` and `@types/json-schema`.

## Type of change

- [x] Bug fix

## Testing

- `uv run pytest utils/tests/` - 329 passed.
- New regression tests: DefinitelyTyped bare-name root, `./package/` root, multi-root and root-level tarballs left untouched, and a precision guard that an injected file inside a bare-name-rooted package is still flagged.
- Against the real commit from #1171, `verify_vendored_node_modules('JamesIves', 'github-pages-deploy-action', 'fa24774553152dd7873cd16ebd8d959b010c5445')` now reports 116 packages verified, 0 extra, 0 mismatched, 0 errors (previously 116 verified plus 13 extra).

---

### DIFF
--- a/utils/tests/verify_action_build/test_npm_registry_verify.py
+++ b/utils/tests/verify_action_build/test_npm_registry_verify.py
@@ -36,12 +36,17 @@ from verify_action_build.npm_registry_verify import (
 )
 
 
-def _make_tgz(files: dict[str, bytes]) -> bytes:
-    """Build an npm-style ``.tgz`` (everything under ``package/``)."""
+def _make_tgz(files: dict[str, bytes], root: str = "package") -> bytes:
+    """Build an npm-style ``.tgz``.
+
+    ``root`` is ``package`` by convention; DefinitelyTyped publishes under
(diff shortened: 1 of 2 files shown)


dfoulks1 (approved):
LGTM

PR #1175, opened 2026-08-12
verify-action-build: recognise aube-lock.yaml as a node lock file

potiuk:
# Code change

## Summary

`jdx/mise-action` v4.2.4 moved its build to [aube](https://github.com/jdx/aube), which writes `aube-lock.yaml` instead of `package-lock.json`. The file is pnpm's lockfileVersion 9 format and carries resolved versions plus sha512 integrity for all 460 transitive deps, so the pinning guarantee is the same one the other node lock files give. The presence check only knew the five established filenames, so #1167 hard-failed with "no matching lock file" on a repo that has one.

- Accept `aube-lock.yaml` in the lock-file presence check and the dependency-pinning note (`security.py`), and exclude it from the source diff as a generated lock file (`diff_source.py`).
- Teach the rebuild to install from it. pnpm reads the file once renamed; `--no-frozen-lockfile` is required because aube records an undeclared direct `tslib` dep that trips pnpm's manifest check. Without this branch the rebuild silently fell through to an unpinned `npm install`.

## Type of change

- [x] Bug fix

## Testing

- `uv run pytest utils/tests/` - 330 passed, including a new lock-file-presence case for `aube-lock.yaml`.
- Against the real commit from #1167, `analyze_lock_files('jdx', 'mise-action', '7e36c90d9ab29c415a2384db3006f3ec8a8cc654')` now reports `node: package.json -> aube-lock.yaml` with no errors.
- Verified pnpm 11 consumes the real `aube-lock.yaml` in a `node:24-slim` container: it validates all 460 entries and installs the lockfile-pinned versions.

---

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -678,6 +678,17 @@ class TestAnalyzeLockFiles:
         }
         assert self._run(files) == []
 
+    def test_node_package_json_with_aube_lock_passes(self):
+        # apache/infrastructure-actions#1167: jdx/mise-action v4.2.4 moved its
+        # build to aube (jdx/aube), which writes aube-lock.yaml in pnpm's
+        # lockfileVersion 9 format — resolved versions plus sha512 integrity
+        # for every transitive dep, so the pinning guarantee is unchanged.
+        files = {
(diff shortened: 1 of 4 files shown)


dfoulks1 (approved):
LGTM

PR #1193, opened 2026-08-21
Update pnpm/action-setup tag in actions.yml

pjfanning:
This one will cause some pain for a few teams as it is used by maybe 10 (P)PMCs.

It highlights how some (P)PMCs do not worry about their github action versions because some of them use v2 and now we have v6.0.10 as most recent version.

https://github.com/search?q=org%3Aapache+pnpm%2Faction-setup&type=code

I'll leave it to the reviewers to decide if we've reached the stage where we need to start actually pushing out stricter version management.

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -834,8 +834,8 @@ phoenix-actions/test-reporting:
   7317eea6e13c47348dd0bb318669485157c518d6:
     tag: v16
 pnpm/action-setup:
-  '*':
-    keep: true
+  0977fd99725f1db4007ccb2928dbb4e90d06cc86:
+    tag: v6.0.10
 posit-dev/setup-air:
   cf390573ff4fe0f198f35df3a642b1409328d859:
     tag: v1.0.1


raboof:
I think this would be a good change - we trust pnpm, but there's a risk of latent issues in old versions, and it's a sign people are not following https://infra.apache.org/github-actions-policy.html .

I did a quick search of remaining projects still on previous versions and pinged them (if they have github issues enabled) with

```
gh issue create --repo "..." --title "Introduce automation to keep \`pnpm/action-setup\` up-to-date" --body $'Hello! This issue was created semi-automatically because this repo popped up in a search. Per https://infra.apache.org/github-actions-policy.html, 3rd-party actions such as `pnpm/action-setup` must be referred to by hash and kept up-to-date by automation such as dependabot. It looks like this repo does not pollow this policy yet.\n\nWe plan to start enforcing this policy for `pnpm/action-setup` through https://github.com/apache/infrastructure-actions/pull/1193 .\n\nIf this was a false posivite, you can close this issue. Otherwise, you likely want to update your workflows, since they will stop working once that PR is merged.'
```

Let's wait a bit to give them a chance to update their workflows and then merge this.

raboof:
(maybe we should turn this into a script, and recommend https://github.com/suzuki-shunsuke/pinact )

raboof (approved):
I think we gave enough notice, this now LGTM

PR #1205, opened 2026-08-27
verify-action-build: diff and scan the shell payload actions ship

potiuk:
# Code change

## Summary

A node action can shell out to a script committed beside it, and nothing was reviewing that script.

- The approved-vs-new source diff copied only `.js/.ts/.mjs/.cjs/.mts/.cts/.json/.yml/.yaml`, so `.sh` files and `Dockerfile*` never appeared in it.
- `analyze_scripts` ran only for composite/docker actions, and only for scripts named in `action.yml` or a `Dockerfile`.

Surfaced by #1196 (`uraimo/run-on-arch-action` v3.2.0). Its `action.yml` declares `main: src/run-on-arch.js`, which `exec()`s `src/run-on-arch.sh` - so the reviewer was shown "only an `action.yml` description changed" while the upstream diff also carried `src/run-on-arch.sh` +12/-1 (five new bind-mounts of runner-writable paths into the container) and three new `Dockerfiles/Dockerfile.*`.

- Add shell/interpreter extensions and `Dockerfile*` to the source diff (`diff_source.py`).
- Discover committed shell scripts from the repo tree, reusing the JS discovery's directory filters (`security.py`).
- Run script analysis for JS actions too (`verification.py`). Informational there: findings show in the summary but are not added to `non_js_warnings`, so a JS action's pass/fail verdict is unchanged.
- README updated.

## Type of change

- [x] Bug fix
- [x] Documentation update

## Testing

- `uv run pytest utils/tests/` - 334 passed.
- New `test_diff_source.py` covering the selection predicate, including suffixed-Dockerfile and not-a-Dockerfile precision cases.
- New `analyze_scripts` regression tests: the run-on-arch shape (script reachable only from the JS entrypoint), vendored/test dir exclusion, sub_path scoping.

---


dfoulks1 (approved):
LGTM

PR #1207, opened 2026-08-27
analyze-action-pr: record the Aug triage precedents and two case-E shapes

potiuk:
# Code change

## Summary

Triaging #1167, #1171, #1196 and #1195 turned up two case-E gap kinds the skill did not describe, and a cheaper way to run step 2.

- **Step 2** - read the `verify` job's CI log instead of re-running the tool locally when the check has already run. Same tool, same output, no Docker rebuild, and it is the only option when the local environment cannot do one. Includes the ANSI-stripping incantation, without which the summary table is unreadable.
- **Case F** - note that not every unverified in-tree binary is F. A bundler-copied dependency asset carries npm's provenance chain rather than the action's, so demanding a GitHub release attestation for it is case E.
- **Case E common shapes** - four new rows: new lock-file format (#1175), registry tarball root assumed (#1174), executable source not diffed or scanned, binary verifiable by rebuild rather than by release.
- **Recent precedents** - one row each for #1167, #1171, #1196 and #1195, each carrying the reusable heuristic it produced (confirm a "missing lock file" verdict by listing the repo root; all-files-of-one-package extra means tarball resolution failed, not tampering; for a node action check the upstream compare API for changed non-JS files; search the upstream repo before drafting an issue).

No behaviour change - documentation only.

## Type of change

- [x] Documentation update

## Testing

- Markdown only; no code paths touched.

---


dfoulks1 (approved):
LGTM

PR #1247, opened 2026-09-07
verify-action-build: exempt test-fixture binaries from in-tree check

potiuk:
## What changed

- `_looks_like_in_tree_binary()` now exempts a binary whose *directory* path carries both a test segment (`test`, `tests`, `spec`, `specs`, `__tests__`) and a fixture segment (`fixture`, `fixtures`, `__fixtures__`), or a `testdata` / `test-data` segment.
- Rationale: a test fixture is test *input*. The action's entrypoint never reaches it, so it never runs on a consumer's runner and cannot be a supply-chain vector. Asking upstream for a SLSA attestation over sample data is the wrong ask. Same reasoning already applied to only-in-rebuild output (#1123).
- Requiring *both* segments keeps the rule narrow: `tests/bin/helper.exe` and a shipped top-level `fixtures/` tree are still flagged.

Unblocks #1218 (anchore/sbom-action v0.24.2), which ships `tests/fixtures/image-debian-match-coverage/java/example-java-app-maven-0.1.0.jar` for SBOM coverage tests and was rejected as an unverifiable in-tree binary.

## Test plan

- New `test_test_fixture_binaries_exempt` covers the exact #1218 path, four other conventional spellings, and three precision cases that must stay flagged.
- `uv run pytest utils/tests/` - 351 passed.

### DIFF
--- a/utils/tests/verify_action_build/test_security.py
+++ b/utils/tests/verify_action_build/test_security.py
@@ -1658,6 +1658,33 @@ class TestLooksLikeInTreeBinary:
             "node_modules/evil-pkg/scripts/externals/7zdec.exe"
         ) is True
 
+    def test_test_fixture_binaries_exempt(self):
+        # Binaries under a test-fixture tree are test *input*, never
+        # reached by the action's entrypoint, so they never run on a
+        # consumer's runner.  anchore/sbom-action@v0.24.2 was
+        # false-flagged for this sample app used in SBOM coverage tests.
+        assert _looks_like_in_tree_binary(
+            "tests/fixtures/image-debian-match-coverage/java/"
+            "example-java-app-maven-0.1.0.jar"
+        ) is False
+        # Other conventional spellings of the same thing.
+        for path in (
+            "test/fixtures/sample.jar",
+            "spec/fixtures/nested/deep/lib.so",
+            "__tests__/__fixtures__/app.exe",
+            "testdata/binary.dll",
+            "src/testdata/golden/tool",
+        ):
+            assert _looks_like_in_tree_binary(path) is False, path
+        # Precision: a test tree alone is not enough — a helper binary
+        # that really is executed still gets flagged.
+        assert _looks_like_in_tree_binary("tests/bin/helper.exe") is True
+        # ...and so is a fixtures tree outside any test directory, which
+        # is a shape real actions use for shipped sample data.
+        assert _looks_like_in_tree_binary("fixtures/payload.jar") is True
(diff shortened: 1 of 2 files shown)


potiuk:
cc @dfoulks1 @ppkarwasz - context for the review, since this one is short but the reasoning behind it matters more than the diff.

`verify-action-build`'s in-tree binary check scans the whole upstream repo. anchore/sbom-action ships a sample Java app as an SBOM test fixture:

```
tests/fixtures/image-debian-match-coverage/java/example-java-app-maven-0.1.0.jar
```

The check flags it as an unverifiable pre-compiled binary and rejects the bump (#1218). That is a false positive, not a finding. The action's entrypoint is `dist/index.js` and it never reaches that jar, so the jar never runs on a consumer's runner and cannot be a supply-chain vector. Asking anchore for a SLSA attestation over test sample data would be the wrong ask.

Same class as #944/#960, where `gradle/wrapper/gradle-wrapper.jar` was false-flagged for the same reason and #951 added a path-suffix exemption. The reasoning also matches #1123: not published means it never runs on a consumer's runner, so it is informational rather than a hard failure.

The rule is deliberately narrow. It requires **both** a test segment (`test`, `tests`, `spec`, `specs`, `__tests__`) and a fixture segment (`fixture`, `fixtures`, `__fixtures__`), or a `testdata` / `test-data` segment. So these are still flagged:

- `tests/bin/helper.exe` - a test tree alone is not enough, and a helper binary there may really be executed
- `fixtures/payload.jar` - a fixtures tree outside any test directory, which is a shape real actions use for shipped sample data
- `dist/test-fixtures.jar` - the segments must be directories, not the filename

The regression test covers the exact #1218 path plus those three precision cases. 351 tests pass, prek clean.

Unblocks #1218, which is otherwise a clean bump.

potiuk:
@raboof @dfoulks1 @ppkarwasz - bumping this one because it now blocks something concrete.

#1218 (anchore/sbom-action 0.24.0 -> 0.24.2) fails `verify` on exactly the false positive this PR fixes:

```
tests/fixtures/image-debian-match-coverage/java/example-java-app-maven-0.1.0.jar:
  no SLSA attestation and release v0.24.2 has no SHA256SUMS
```

Everything else on #1218 is green. Once this lands, a rebase should clear `verify` and the bump can go in.

All 8 checks green here. The diff is ~30 lines of exemption logic plus its regression test.

dave2wave (approved):
LGTM

PR #1248, opened 2026-09-08
stash/restore: back off between download retries

potiuk:
# Code change

## Summary

`stash/restore` retried a failed `gh run download` immediately, so every
attempt landed inside the same congestion window that caused the first
failure. When many jobs of one matrix pull the same large stash at once,
the artifact backend resets connections mid-stream — and the retry loop
re-synchronised the herd instead of breaking it up.

Seen on `apache/airflow` run 34200137532: 6 of the 14 Kubernetes jobs that
started within the same 90 seconds burned all three attempts on
`read: connection reset by peer` in under two minutes, while the 20 jobs of
the earlier wave restored the same artifact without trouble. The stash was
found and intact in every case.

- Adds a `retry-delay` input (default `5`) — base seconds for the wait
  between download attempts.
- Waits with exponential growth and **full jitter**, capped at 60s per
  sleep. Full jitter is the point: spacing the retries evenly would keep
  the failing jobs in lockstep, randomising the whole interval pulls them
  apart.
- No wait after the final attempt, and none for non-transient exit codes.
- `retry-delay: "0"` restores the previous immediate-retry behaviour.

## Type of change

- [X] Bug fix
- [X] Enhancement to existing code

## Testing

`stash/restore/test_download_stash.py` — 15 tests, all passing. The
`sleep`/`rand` hooks are injected so the suite asserts the computed waits
without sleeping (runs in 0.005s). New cases cover the backoff sequence,
jitter scaling, the 60s cap, no-wait-after-last-attempt, no-wait on
non-transient exits, and both the `"0"` and unset defaults.

`ruff check` reports the same 6 pre-existing findings before and after —
no new ones.

---
Drafted-by: Claude Opus 5; reviewed by @potiuk before posting

### DIFF
--- a/stash/restore/action.yml
+++ b/stash/restore/action.yml
@@ -52,6 +52,12 @@ inputs:
       the download will be retried until it succeeds or this many attempts have been
       made. Other exit codes are not retried.
     default: "3"
+  retry-delay:
+    description: >
+      Base number of seconds to wait between download attempts. The wait grows
+      exponentially with each attempt and is fully jittered, so jobs that fail
+      together do not retry together. Set to "0" to retry immediately.
+    default: "5"
   fail-on-download:
     description: >
       If true, the action will fail when a stash artifact was found but could not be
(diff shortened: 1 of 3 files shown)


dfoulks1 (approved):
LGTM

PR #1259, opened 2026-09-09
utils: add org-wide queued/running Actions job snapshot

potiuk:
# Code change

## Summary

Adds `utils/actions-queue-status.py`, which reports every repository in the org that has GitHub Actions jobs queued or running right now.

Debugging a slow ASF runner queue currently means guessing. `GET /orgs/apache/actions/runners` reports each runner's `status` and `busy` flag but needs `admin:org`, and there is no org-wide REST endpoint for queued jobs at all. Check-run state is readable by anyone who can read the repository and maps one-to-one onto workflow jobs, so this works without Infra privileges.

- **Discovery** reads the workflows directory straight out of the git tree (`object(expression: "HEAD:.github/workflows")`), so a repo counts as using Actions only when it really has workflow files — no `pushedAt` heuristics, no per-repo REST probe. 3,186 repos scanned, 1,256 use Actions.
- **Status** batches repos into aliased GraphQL queries (~20 repos per request) and counts check runs. Output is two tables — sorted by running jobs and by queued jobs — plus `--csv` (writes `-by-running.csv` and `-by-queued.csv`) and `--json`.
- **Each API where it is strongest.** A sampled count is only trustworthy if the sample provably covered the repo, so every query returns two `totalCount`s: open PRs, and check suites per commit. A repo is trusted to GraphQL only when *both* limits held; everything else is re-counted over REST, which is exact per repo. In practice ~8 of the ~160 active repos are thin enough for GraphQL alone — the cheap pass is a filter, and anything with real activity earns an exact count.
- **Rate-limit safety:** every query requests `rateLimit { cost remaining resetAt }` and the sweep stops at 200 points remaining, reporting partial counts rather than dying; 502s and rate-limit rejections retry with 4s/16s/64s backoff; `--workers` defaults to 3 because 6 trips GitHub's per-minute points cap.

Two things worth recording for anyone else writing against this API:

- The REST `/rate_limit` endpoint is not a usable pre-flight check. It reported `graphql: 5000/5000, used=0` while the API was actively rejecting queries with `RATE_LIMIT`. Only the `rateLimit` block returned inside each query is trustworthy, so that is what the script tracks.
- GraphQL's node cost climbs sharply with `--prs`: measured across the 1,256 repos, a sweep costs ~250 points at `--prs 3`, ~2,700 at 5 and ~21,000 at 10, against 5,000 per hour. Since raising it only covers *quiet* repos, and each uncovered repo costs about one cheap REST call, the low default is both cheaper and no less accurate. Discovery prints the open-PR distribution each run so the trade-off can be re-checked.

Documented in `README.md` under "Snapshotting Queued and Running Actions Jobs".

## Type of change

- [x] New utility or action
- [ ] Bug fix
- [ ] Enhancement to existing code
- [ ] Workflow or CI change
- [x] Documentation update
- [ ] Other

## Testing

Validated against the live `apache` org, with a full REST sweep as ground truth.

- `prek run --all-files` passes.
- Full-org sweep completes end to end: 3,186 repos scanned, 1,256 with workflows, 63 batched status queries.
- **Paired against ground truth.** A hybrid sweep and a blanket REST sweep of all 1,256 repos, run back to back: 160 vs 156 active repos, disagreeing on 8 repos in one direction and 12 in the other — all low-count, and within the churn of the twenty minutes separating the passes.
- **The coverage gate was caught by that comparison and fixed.** With only the open-PR test, `apache/skywalking-java` (one open PR, thirteen active runs on its head commit) was reported as 32 queued jobs against a true 203. Adding the check-suite `totalCount` routes it to REST; the repos re-counted over REST rose from 674 to 892, and the repos GraphQL is trusted for fell from 22 to 8.
- **Failure paths were exercised for real, not simulated:** an HTTP 502 mid-discovery (now retried, and fatal rather than silently partial), and points exhaustion mid-status (now guarded).

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ This repository hosts GitHub Actions developed by the ASF community and approved
   - [Automatic Expiration of Old Versions](#automatic-expiration-of-old-versions)
   - [Removing a Version](#removing-a-version-manually)
 - [Auditing Repositories for Actions Security Tooling](#auditing-repositories-for-actions-security-tooling)
+- [Snapshotting Queued and Running Actions Jobs](#snapshotting-queued-and-running-actions-jobs)
 
 ## Checking the Action Usage in an ASF Project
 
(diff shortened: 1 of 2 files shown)


dave2wave (approved):
Did you show this output in slack?

potiuk:
> Did you show this output in slack?

Yes.

PR #1273, opened 2026-09-13
utils: colorize CLI help and borrow the gh CLI's token

potiuk:
# Code change

## Summary

`rich-argparse` for the four argparse CLIs in `utils/`, plus a `gh auth token`
fallback so anyone already running `gh auth login` does not have to mint a
second PAT.

- **Colorized help** — `RichHelpFormatter` on `actions-queue-status.py`,
  `actions-audit.py` and `verify-action-build`; `RawDescriptionRichHelpFormatter`
  on `audit-actions-cache.py` (it has a literal-formatted epilog). Declared in
  both `pyproject.toml`s, both PEP 723 headers and the root dev group.
- **Token fallback, scoped rather than unconditional** — in `actions-audit` and
  `verify-action-build`, `GitHubClient` switches to `requests` as soon as a token
  is set, so always filling one in would have silently taken the `gh` CLI out of
  the default path. There the fallback only fires under `--no-gh`, where a token
  is genuinely required. `actions-queue-status` keys its client off `--no-gh`
  rather than token presence, so it chains the full
  `--github-token` -> `GH_TOKEN`/`GITHUB_TOKEN` -> `gh auth token`.
- **`verify-action-build` also primes `GITHUB_TOKEN` in the environment** — its
  lockfile discovery and in-tree binary checks call `api.github.com` directly and
  otherwise share the 60-requests/hour unauthenticated budget, which a single run
  can exhaust.
- **Progress bars for the long loops in `actions-queue-status`** — it spends
  minutes in three loops and said very little while doing it: discovery printed
  one line per 100-repo page, the status sweep one line per ten batches, and the
  REST re-count nothing at all. Each phase now drives a rich progress bar
  (spinner, bar, N/total, elapsed, plus a dim note with repos found so far and
  GraphQL points left). Discovery gets a real total because the page query now
  asks for `repositories.totalCount`. The bars use the existing stderr console
  and are transient, so `--json` on stdout stays a clean document; the note
  column is last and width-capped, so a narrow terminal ellipsizes the
  commentary rather than truncating the counts.
- **Docs** — README `--no-gh` section, both flag tables and both prerequisites
  lists, and a note on the progress output. `audit-actions-cache.py` also gained
  the PEP 723 header it was missing, so its usage examples moved from `python`
  to `uv run`.

## Type of change

- [ ] New utility or action
- [ ] Bug fix
- [ ] Workflow or CI change
- [x] Documentation update
- [ ] Other

## Testing

- `uv run pytest` - 628 passed, 5 skipped (6 new tests covering `gh_auth_token`
  and the `--no-gh` fallback).
- `--help` renders under `rich-argparse` on all four CLIs; the fallback returns a
  live token from each with `GH_TOKEN`/`GITHUB_TOKEN` unset.
- Progress bars exercised end to end: full `apache` discovery (1258 repos) and a
  complete run against a smaller org (88 scanned, 55 with workflows, 3 batches,
  25 REST re-counts). Rendering checked at 80 and 120 columns - at 80 the note
  ellipsizes while the bar and counts stay intact.

Note: no `click` anywhere in the repo, so `rich-click` was not needed - worth
reaching for if any of these ever move to click.

---

https://claude.ai/code/session_01PpWDqP6JqGZ5FhJdrBsoVT


dave2wave (commented):
This a very large change. I scanned through the changes. I noticed several different python scripts using essentially the same code about accessing the GH Token and wonder if that code should be refactored to come from a shared module.

potiuk:
Gentle nudge on this one, @dfoulks1 @ppkarwasz - the ASF Actions queue is looking very long again, and these are exactly the small fixes that make the tool we use to look at it more robust: stored repo list so a sweep does not re-discover 1200+ repos every time, progress output so a long sweep is not a black box, PMC grouping and totals so the snapshot is readable.

Low risk - `utils/` tooling only, no workflow or action surface. CI is green.

Note these two overlap (both add `utils/apache-actions-repos.txt` and touch `actions-queue-status.py`), so whichever lands first will need the other rebased on top - happy to do that as soon as one goes in.

ppkarwasz (approved):
Empirically it works, thanks! :100:  I didn't have time to go through the code line by line.

PR #1276, opened 2026-09-14
Add carabiner v1.2.7 tag with expiration dates

pjfanning:
Example failed build
https://github.com/apache/pekko-connectors-kafka/actions/runs/34817759287/job/103892071090?pr=657

setup-sbt 1.5.9 (latest release) seems to switch to carabiner 1.2.7 (not the latest but close to it)
https://github.com/carabiner-dev/actions/releases


ppkarwasz (commented):
These changes will not be enough to white-list `carabiner-dev/actions`: the actions rely on *sibling* actions, which until recently were pinned by SHA1. Of course this means, each action must pin a **previous** commit of its sibling.

In carabiner-dev/actions#128 I proposed to change those references with the *newly introduced* `$/` syntax. Since the PR was accepted, the problem will solve itself in the next release.

Until then I would replace all those tags with `*`. Ideally I would also replace all the `carabiner-dev/actions/<foo>` entries with `carabiner-dev/actions/*`. Could you check if that works?

pjfanning:
@ppkarwasz the carabiner actions are not used by Pekko directly. We get the transitively from setup-sbt action.

We don't need to upgrade setup-sbt urgently - all we have is some broken Dependabot PRs.
We can ignore those broken PRs and when a future setup-sbt release includes the upgraded carabiner releases, we can see if that resolves the issues.

ppkarwasz:
> @ppkarwasz the carabiner actions are not used by Pekko directly. We get the transitively from setup-sbt action.

Yes, I understood that.

> We don't need to upgrade setup-sbt urgently - all we have is some broken Dependabot PRs.

Let's wait for the next release then. Those actions are the real **trust roots** for `setup-sbt` and similar. It would be a little bit scary to unpin them, even temporarily.

potiuk:
@pjfanning - yeah it's been plaguing us for quite some time. What I can propose is to just use `@dependabot` commands to skip upgrades to the action - you can - I believe say "skip this upgrade of carabiner" - and it will remmember it and will not attempt to upgrade it unti the **next** version that will likely - as @ppkarwasz explained likely have the problem solved thanks to Piotr's PR.

PR #1294, opened 2026-09-16
sweep-open-prs: skill for triaging the whole open PR queue

potiuk:
Adds a `sweep-open-prs` skill for the pass over the whole open PR queue: gather state for every open PR, bucket each one, act only on what is confirmed.

- Two `gh pr list` calls cover the queue - one for state, one for a check rollup collapsed to only non-green results. Avoids the per-PR loop that trips `gh`'s spurious 401.
- Records the judgement rules that decide most of the queue: `BLOCKED` + `REVIEW_REQUIRED` means "needs one approval" rather than "broken"; a self-authored PR can never be self-approved; carried-over warnings are not regressions; a held bump is often superseded rather than fixed.
- Routes anything red, or any bump being approved on substance rather than a green tick, to `analyze-action-pr` instead of duplicating its failure taxonomy.
- Documents both skills in `README.md` under a new **Agent Skills** section, with a TOC entry. `analyze-action-pr` had never been documented; it is included rather than left invisible now that the section exists.

Test plan: `prek run --files README.md .claude/skills/sweep-open-prs/SKILL.md` passes. Frontmatter is 363 chars against the 1024 limit, with the two required fields. Both README links resolve to files in the tree and the `#batch-reviewing-dependabot-prs` anchor matches a real heading. Every `gh` invocation in the skill was run against this repo's live queue while drafting it.


potiuk:
This one adds a SKILL - reflecting what I have been doing recently - swiping through all PRs with my agent and acting appropriately.

With this skill **ANYONYE** can do what I did regularly by just asking the agent "

"sweep throuh open PRs". 

That's it - everything will happen automatically then, you will be asked to confirm the actions when they are safe, or asked to additionally review the PRs when there are doubts.

This should allow anyone to spend few minutes (!) a day - to keep the repo in a good shape.

@dfoulks1 @dave2wave @ppkarwasz -> I would love if you would like to try it and see how much time it saves.

potiuk:
Also @raboof -> With that SKIL I was able to sweep through 14 open PRs in lest than 10 minutes of elapsed time, with analysis of what they do, checking if they are sound, if they have no suspicious changes (this all on top of the deterministic analysis that we run in the PRs).

I **highly recommend** merging that one and starting using it by everyone from time to time. That will help us to avoid SPOFs and delay.

I even have an idea that when we have access to `llmao` we could add a workflow that will be using it automatically and all we get is an "review and approve" buttong to marge result of such a swipe.

This would be fantastic workflow - we would just get a single place where regularly. we will see:

* there are 10 new PRs
* 5 of them are ready to merge, save, checked for malicious code
* 2 of them need attemtion
* 5 of them are regular dependabot dependency upgrades with cooldown that looks safe

And any of us would be able to "Approve" all of them.

PR #1263, opened 2026-09-10
actions-queue-status: colour progress, totals and PMC grouping

potiuk:
A full org sweep of `actions-queue-status.py` runs for minutes behind a handful of terse lines, which makes a slow or partial run hard to explain after the fact. This adds coloured progress and diagnostics.

## What changed

- **`Reporter`** owns everything the script prints: the colour vocabulary (cyan phases, dim detail, green success, yellow warnings, red errors), the progress bars, and thread-safe run counters. **`Tracker`** is its per-phase handle - a live `rich` bar on a terminal, one line every N steps when piped, so logs stay greppable.
- **Phase banners**: Discovery / Status / REST re-count, numbered against the phases the run actually plans (`Phase 2/3`), each closing with elapsed time and GraphQL points spent. When the REST phase is skipped because the sample covered everything, it says so instead of silently vanishing.
- **Bars** show count, percent (speed for the indeterminate discovery pass), a live detail field, elapsed and ETA. The points balance is a traffic light - green above 2000, yellow down to the 200-point floor, red below - so an approaching budget stop is visible before it happens.
- **`Run diagnostics` table** at the end: queries, retries, give-ups, batches split, repos skipped, REST requests, repos re-counted, points spent, wall time. Failure counters stay dim at zero and colour when they are not.
- **New flags**: `-v/--verbose` (retries with backoff and error, batch splits, discovery cursors, each REST re-count next to what GraphQL sampled), `-q/--quiet`, `--no-color` (also honours `NO_COLOR`, for the bars too).
- Everything goes to **stderr**, so `--json` on stdout stays machine-readable.
- The open-PR distribution is now a table with the knee of the curve highlighted; result tables gained a `showing top N of M` caption and a coloured `source` column.

## Drive-by fix

Writing the retry diagnostics surfaced a real defect: both retry loops slept their full backoff *after* the final attempt - up to 256s of dead wait before giving up. They now break out instead, which also makes the new "retrying in Ns" message truthful.

## Verification

- An offline harness with canned GitHub payloads exercised discovery paging, a forced 502 -> retry -> batch split -> skipped repo, the REST re-count, and the piped / TTY / verbose / quiet / JSON / no-color paths.
- A live `--repos-file` run against `apache/airflow` (37 running jobs, re-counted over REST).
- `ruff check` clean.

README is updated in the same commit: the three new flags in the options table plus a "Progress and Diagnostics" section with sample output.

Example, mid-sweep:

```text
[...]

Issue #950, opened 2026-06-20
Should `approved_patterns.yml` use the correct word casing for actions ?


jbampton:
On Sedona SpatialBench we are getting a failure on the ASF Allowlist workflow for `Swatinem/rust-cache`:

https://github.com/apache/sedona-spatialbench/actions/runs/27872942501/job/82487683539?pr=114

We use the correct case `Swatinem` but `approved_patterns.yml` uses `swatinem`. 

https://github.com/Swatinem

https://github.com/Swatinem/rust-cache

https://github.com/apache/infrastructure-actions/blob/25ab499ef9c241b64a56860245e57c195baddec6/approved_patterns.yml#L332

raboof:
Huh, I'm surprised that's case-sensitive - but if it is, we should probably use the 'correct' case indeed.

PR #1297, opened 2026-09-16
verify-action-build: credit vendored files git normalised from CRLF

potiuk:
`Vendored npm registry check` byte-compares each committed file against the integrity-verified registry tarball. A package whose tarball ships CRLF can never pass that comparison: git normalises line endings on commit, so the committed blob holds the LF form while the tarball does not. Every such file was reported as `modified`, which reads as tampering.

- Retries a failed byte comparison against the tarball content with `\r\n` folded to `\n`, and credits the file when that hashes to the committed blob.
- Reports those files in a new `crlf_normalized` category rather than silently folding them into `verified`, so the output says why the file did not byte-match.
- Adds a regression test for the shape, plus a precision guard asserting a real content edit still fails after the folding is in place.

This is a proof rather than a heuristic: if the folded bytes hash to the committed blob, the committed file *is* the published content, line endings aside.

**Live case — #1283.** `benchmark-action/github-action-benchmark` vendors `node-fetch`'s nested `whatwg-url@5.0.0`, a 2016-era package whose tarball ships `lib/url-state-machine.js` with CRLF:

| | line endings | CR count | bytes | git blob |
| --- | --- | --- | --- | --- |
| registry tarball | CRLF | 1297 | 33573 | `c25dbc2c4862...` |
| committed in the action | LF | 0 | 32276 | `27d977a25f90...` |

`33573 - 32276 = 1297`, exactly the CR count, and `diff --strip-trailing-cr` reports the two identical. Folding the tarball's CRLF hashes to `27d977a25f9011fc31ed28c17ba11f729b85edea` - the committed blob GitHub reports. The other 26 vendored packages in that action matched byte-for-byte, so this was a single file failing for a reason unrelated to the action's security posture, on every bump.

Test plan: `uv run pytest utils/tests/` - 353 passed. `prek run --files ...` passes. The two new tests fail without the fix (reproducing #1283's exact output: `content differs from registry tarball`) and pass with it.


potiuk:
Nudge on my open PRs, @dfoulks1 @ppkarwasz - five are green and waiting, and they are easier to read as one set than separately.

Two are checker fixes, and these are the ones worth your time:

- **#1297** - `verify-action-build` reported a vendored file as modified whenever the npm tarball shipped CRLF, because git normalises line endings on commit and the committed blob can then never byte-match. That is what made #1283 red. Without it the same false positive returns on every future bump of that action.
- **#1305** (closes #1304) - `fnmatch` applies `os.path.normcase`, the identity on POSIX, so a correctly-cased `Swatinem/rust-cache@v2` was rejected against a lower-cased entry. apache/sedona-spatialbench hit it (#950); #1300 fixed the single entry, this removes the class. Only `owner/repo` folds - tags and SHAs stay case-sensitive, with tests pinning that.

Two record how the checkers get used:

- **#1294** - a `sweep-open-prs` skill for triaging the whole open PR queue in one pass, plus a README section documenting both skills.
- **#1303** - corrects the `analyze-action-pr` precedent row that still says carabiner-dev's installers curl without verification. That upstream issue is closed and those actions now verify SLSA provenance by default, so the row was steering reviewers to re-raise a settled finding. Adds three precedents.

And one tool change:

- **#1263** - `actions-queue-status` colour, totals and PMC grouping. Rebased onto current main after #1273 landed; the two had built competing progress-reporting designs, so this keeps the `Reporter` abstraction and main's stored-repo-list behaviour.

No rush on any of them, and happy to split #1263 up if the rebase makes it awkward to review. #1297 is the one I would land first if you only have time for one.

raboof (approved):
ugh

PR #1298, opened 2026-09-17
allowlist: add Cargo-Rail-Action v10.0.0

loadingalias:
Apache Iggy needs the Cargo-Rail planner for https://github.com/apache/iggy/pull/3999. The GHA installs Cargo-Rail and produces a validated named-work plan for CI routing. Planning failures restore Iggy's full workspace or all image fallback.

This request covers only the root planner Action. It does not request approval for the repository's nested setup, cache, reporting, or release Actions.

**Name of action:** Cargo-Rail planner

**URL of action:** https://github.com/loadingalias/cargo-rail-action

**Version to pin to ([hash only](https://infra.apache.org/github-actions-policy.html)):**`e706767c343acb6a2ab1917f243b1616ae9eb05b`

## Permissions

The planner requires only `contents: read`. Its `repository-token` input defaults to `github.token` and is only used when missing Git history must be fetched from the same repo.

The token is not placed in URLs, command arguments, Git config, Action outputs, summaries, or unrelated child processes. The planner requires no repo write permission or OIDC token and has no post step.

## Related Actions

No currently approved Action provides Cargo-graph-aware named-work planning w/ validated Cargo, target, and variant selectors.

Installing Cargo-Rail directly supplies the CLI, but it does not provide GitHub-event comparison selection, bounded history acquisition, independent plan validation, or bounded GitHub Action outputs.

## Security review

Cargo-Rail Action v10.0.0 and its default Cargo-Rail v0.28.1 release are immutable and have GitHub build-provenance attestations:

- https://github.com/loadingalias/cargo-rail-action/releases/tag/v10.0.0
- https://github.com/loadingalias/cargo-rail/releases/tag/v0.28.1

The composite bootstrap downloads its release manifest and platform runtime over HTTPS, enforces size bounds, verifies the runtime SHA-256 digest, and runs a version and target self-check. The runtime installs the selected Cargo-Rail archive and verifies it against that release's `SHA256SUMS`.

The installer does not verify attestations at runtime. Release immutability and the attestations above are the reviewed publication trust anchors.

## Checklist

- [ ] Compiled JavaScript in `dist/` matches a clean rebuild (verify with `uv run utils/verify-action-build.py org/repo@hash`)
---
Drafted: @loadingalias; Automated Review by Codex (Sol-xhigh)


potiuk:
Thanks for the thorough submission, @loadingalias - the pin is correct (`v10.0.0` is an annotated tag and dereferences to exactly `e706767c343acb6a2ab1917f243b1616ae9eb05b`), the release history is consistent, the repo is MIT licensed, and `verify` passes cleanly.

Holding on one point that `verify` cannot see. The Action is composite and its payload is `scripts/bootstrap.sh`, which downloads the Cargo-Rail runtime from the GitHub release and checks it against `cargo-rail-action-runtime-v1.tsv` fetched from that same release. That is integrity but not authenticity: whoever can write release assets can replace the binary and the manifest together, and every check in the script still passes. `release.yml` has no `id-token: write` and no signing step, and the v10.0.0 assets carry no signature or attestation.

The script itself is careful work - HTTPS-only, size-capped, strict manifest validation, tight permissions, and a `runtime-source: source` escape hatch - so this is one gap rather than a general concern, and `RUNTIME_VERSION` comes from `Cargo.toml` in the pinned tree, so the release is fixed by the commit rather than floating.

Filed upstream as https://github.com/loadingalias/cargo-rail-action/issues/3, proposing `actions/attest-build-provenance` so consumers can `gh attestation verify`. We have taken the same position before and upstreams have generally accepted it quickly (see jdx/mise-action#547, fixed the same day). Once the assets are attested, this should go through straightforwardly.

potiuk:
Thanks for the detailed writeup, and for scoping the request to just the root planner Action.

I went through the supply-chain side and it holds up well:

- The v10.0.0 release is immutable, so the assets cannot be swapped under the pinned SHA.
- Build-provenance attestations are present for the runtime binaries.
- `scripts/bootstrap.sh` bounds the manifest and runtime sizes, verifies the runtime SHA-256 against the manifest, and runs a version/target self-check before exec.
- MIT LICENSE and `Cargo.lock` are present, and `verify` is green (one soft warning: `repository-token` defaults to `github.token`).

What I cannot get past yet is not security, it is demand and maintenance surface:

1. `loadingalias` is the only human contributor to cargo-rail-action (the remaining commits are the bot), on a repo created in December 2025 with 4 stars and 0 forks. That is the "sufficient number of contributors or has contributors within the ASF community" checklist item, which is the one left unchecked in the description.
2. apache/iggy#3999, the consuming PR, is still a draft and behind main, with no review from an Iggy committer. So no ASF project has committed to taking this dependency yet.

Allowlisting an action commits the whole foundation to it, so I would like the Iggy side settled first. Concretely: get #3999 out of draft and approved by an Iggy committer, link that here, and I will take another pass. If the Iggy maintainers want the planner, the rest of this request is in good shape.

---
Drafted-by: Claude Opus 5; reviewed by @potiuk before posting

potiuk (approved):

potiuk:
The upstream action fix was merged. So ... we can merge this one, next version will have provenance.

loadingalias:
Thank you, @potiuk. I really appreciate you taking the time. It directly improved the Action and its release process. I'm sorry you beat me to it, but thankful you did.

I've updated the Iggy PR to v10.0.1 at the exact allowlisted SHA.

potiuk:
Cool :) . Glad I could help.

PR #1300, opened 2026-09-17
fix(actions): correct casing for owner identifier of approved` rust-cache` action

### LINKED ISSUE #950: Should `approved_patterns.yml` use the correct word casing for actions ?

hainenber:
Closes https://github.com/apache/infrastructure-actions/issues/950

Not sure if there should be any follow-up acts to prevent similar cases, such as having a sample repo that utilizes all approved GHAs for E2E? :D


potiuk (approved):
Confirmed the mechanism rather than just the description: `check_asf_allowlist.py:156` matches with `fnmatch.fnmatch`, which applies `os.path.normcase` - identity on POSIX. On the Linux runners `fnmatch("Swatinem/rust-cache@v2", "swatinem/rust-cache@*")` is `False`, so the lowercase entry really did reject correctly-cased usage. The owner is `Swatinem`, and #950 is a real consumer hitting it (apache/sedona-spatialbench).

Touching only `actions.yml` is the complete fix, despite #950 naming `approved_patterns.yml`: that file carries `# This file was generated from actions.yml by gateway/gateway.py. It will be regenerated and committed as part of various workflows.` so `actions.yml` is the single source and the sync regenerates the rest.

On the follow-up you raise: worth doing, and the general form is to make the matching case-insensitive rather than to police the casing of every entry, since GitHub owner and repo identifiers resolve case-insensitively. Opened as https://github.com/apache/infrastructure-actions/issues/1304. Out of scope for this PR.

LGTM.

